PR #3, opened 2022-09-17
feat: Security Spec

HarryET:
# Description
This PR adds the security spec for communication between the Relay and Echo Server instances. This allow's both the Relay and Echo Server to validate requests they have received.

There will potentially be a future change to the spec for Client --> Echo Server auth but that is not included in this PR.

## Due Diligence

* [x] Breaking change

### DIFF
--- a/spec/auth.md
+++ b/spec/auth.md
@@ -2,16 +2,16 @@
 
 ## Relay <--> Echo Server
 
-We use Ed25519 to verify that requests from the Relay network are valid and intended for the Echo Server Instance that has received them.
+We use Ed25519 to verify that requests from the Relay are valid and intended for the Echo Server Instance that has received them.
 
 ### Implementation
 
 Echo Server generates<sup>[[1]](#generating-keys)</sup> an Ed25519 key-pair and stores that in-memory - or in files specified using environment variables. The generated
-public key is then sent to the Relay network along with a cloud app project id and the public url (both from environment variables) so that
-future requests to the relay can be signed and validated. The Echo Server instance also fetches the Relay network's public key and caches it
+public key is then sent to the Relay along with a cloud app project id and the public url (both from environment variables) so that
+future requests to the relay can be signed and validated. The Echo Server instance also fetches the Relay's public key and caches it
 so that requests received can be validated.
 
-On start-up Echo Server sends [the register request](#post-pushservers) to the relay network<sup>[[2]](#registering-with-relay)</sup>. This request is formatted using the
+On start-up Echo Server sends [the register request](#post-pushservers) to the relay<sup>[[2]](#registering-with-relay)</sup>. This request is formatted using the
 public key from the above key-pair as well as the provided `projectId` and `publicUrl` (both provided via Environment Variables).
 
 ### Relay Endpoints
@@ -19,7 +19,7 @@ public key from the above key-pair as well as the provided `projectId` and `publ
 #### GET `/public-key`
 ##### Response
 > **Note**
-> This is an example Ed25519 Public Key, this is not a valid Public Key for the Relay Network.
+> This is an example Ed25519 Public Key, this is not a valid Public Key for the Relay.
 ```
 693a98827a9c7e8f818af53b9720671eb4d3075815a8c2c8f6d0da12ba1aba7a
 ```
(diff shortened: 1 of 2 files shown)


spec/auth.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+#### POST `/push/servers`
+##### Request
+> **Note**
+> This is an example request. The `project_id` & `public_key` values are not valid.
+```json
+{
+    "project_id": "83f11e753439fab08222b45e2d029eab",
+    "public_key": "d5aa4b55ecf4553c3ef8f8a945d9449394f0b3b7787af049d1d4828037465a4f",
```
arein: should this be a header?

HarryET: no, as this is a request to the Relay's to register this public key, project id and url combination

spec/auth.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Implementation
+
+Echo Server generates<sup>[[1]](#generating-keys)</sup> an Ed25519 key-pair and stores that in-memory - or in files specified using environment variables. The generated
+public key is then sent to the Relay network along with a cloud app project id and the public url (both from environment variables) so that
+future requests to the relay can be signed and validated. The Echo Server instance also fetches the Relay network's public key and caches it
+so that requests received can be validated.
+
+On start-up Echo Server sends [the register request](#post-pushservers) to the relay network<sup>[[2]](#registering-with-relay)</sup>. This request is formatted using the
+public key from the above key-pair as well as the provided `project_id` and `public_url` (both provided via Environment Variables).
```
sbc64: We need to maintain consistency in the casing of the variables.

`project_id` should be `projectId`: https://github.com/WalletConnect/rs-relay/pull/354/files#diff-514195183eed7f300b70bc7617a8daf388bfdd6d01a52e3db2900aec6f66e4abR83

`public_url` is a new variable, but we should keep the same format of camelCase: `publicUrl`

HarryET: :+1: I will change it to use camelCase throughout

sbc64: It doesn't matter, but I think we should avoid fragmentation right now.

HarryET: I'm happy using camelCase I just defaulted to using underscores as I was unsure what is currently used throughout

spec/auth.md
```diff
@@ -0,0 +1,59 @@
+# Echo Server Authentication
+
+## Relay <--> Echo Server
+
+We use Ed25519 to verify that requests from the Relay network are valid and intended for the Echo Server Instance that has received them.
+
+### Implementation
+
+Echo Server generates<sup>[[1]](#generating-keys)</sup> an Ed25519 key-pair and stores that in-memory - or in files specified using environment variables. The generated
+public key is then sent to the Relay network along with a cloud app project id and the public url (both from environment variables) so that
```
sbc64: `is then send to the Relay Network`.

The Relay network doesn't exist yet. Can you clarify who the consumers will be of this public key ?

HarryET: I'll reword to make it make sense. Its actually just the rust relay...

sbc64: From the proposal: it seems that it should be the all the relay servers that query the public key: https://docs.google.com/document/d/1DPmAea-kC7wWVgj5H6eKFVpYzRXCUk8FvhosEMt0aK0/edit#

HarryET: Yeah, thats correct I have changed the .md file to reflect that too

arein (commented):
Let's close on this on Monday

spec/auth.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Relay Endpoints
+
+### GET `/public-key`
+#### Response
+> **Note**
+> This is an example Ed25519 Public Key, this is not a valid Public Key for the Relay.
+```
+693a98827a9c7e8f818af53b9720671eb4d3075815a8c2c8f6d0da12ba1aba7a
+```
```
arein, line 28: this works for now. Later we can consider pinning it in ENS or `relay.walletconnect.com/.webhook_pk`

arein (approved):

spec/auth.md
HarryET: :+1:

PR #4, opened 2022-09-17
feat: Migrate to axum

### LINKED ISSUE #2: Metric Dimension Cardinality too High
https://github.com/WalletConnect/echo-server/blob/main/src/handlers/register_client.rs#L80

https://www.robustperception.io/cardinality-is-key/

Cardinality should be max 10.

HarryET:
# Description
- Migrate to axum
- Fix responses
- Fix metrics, fixes #2 
- Switch to Async Traits

## How Has This Been Tested?
Unit Tests

## Due Dilligence

* [x] Breaking change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ build = "build.rs"
 
 [dependencies]
 tokio = { version = "1", features = ["full"] }
-warp = "0.3"
+axum = { version = "0.6.0-rc.2", features = ["json"] }
 
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
@@ -27,6 +27,8 @@ tracing-opentelemetry = "0.17.4"
 
 a2 = { git = "https://github.com/walletconnect/a2.git", branch = "v0.7" }
 fcm = "0.9.2"
+
+async-trait = "0.1.57"
 
 [dev-dependencies]
 serial_test = "0.9.0"
--- a/src/error.rs
+++ b/src/error.rs
@@ -8,6 +8,7 @@ pub enum Error {
     TraceError(opentelemetry::trace::TraceError),
     MetricsError(opentelemetry::metrics::MetricsError),
     ApnsError(a2::Error),
+    FcmError(fcm::FcmError),
     IoError(std::io::Error),
     ProviderNotFound(String),
     ProviderNotAvailable(String),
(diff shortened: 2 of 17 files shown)


arein (approved):

Cargo.toml
```diff
@@ -9,7 +9,7 @@ build = "build.rs"
 
 [dependencies]
 tokio = { version = "1", features = ["full"] }
-warp = "0.3"
+axum = { version = "0.6.0-rc.2", features = ["json"] }
```
heilhead, line 12: Why pre-release version?

src/error.rs
```diff
@@ -59,6 +61,12 @@ impl From<a2::Error> for Error {
     }
 }
 
+impl From<fcm::FcmError> for Error {
```
heilhead, line 64: I'd suggest using [`thiserror`](https://crates.io/crates/thiserror) for `Error` and its `#[from]` macro for automatic `From` implementations.

src/handlers/delete_client.rs
```diff
@@ -1,49 +1,46 @@
...
 use std::sync::Arc;
-use warp::http;
 
 pub async fn handler(
-    id: String,
-    state: Arc<State<impl crate::store::ClientStore>>,
-) -> Result<impl warp::Reply, warp::Rejection> {
+    Path(id): Path<String>,
+    State(state): State<Arc<AppState<impl crate::store::ClientStore>>>,
+) -> impl IntoResponse {
```
heilhead, line 13: I'd suggest refactoring the return type to something like `Result<(), HandlerError>` and then implement `impl<T: Serialize> IntoResponse for Result<T, HandlerError>` which would serialize the data into JSON and specify the status code. Doing this would allow you to handle errors naturally using `?` inside your handler functions, and clean up the JSON serialization.

src/providers/noop_test.rs
```diff
@@ -33,9 +33,7 @@ mod noop_test {
     fn send_notification() {
         let mut provider = NoopProvider::new();
 
-        provider
-            .send_notification(PUSH_TOKEN.to_string(), EXAMPLE_MESSAGE.to_string())
-            .expect("Failed to send notification");
+        provider.send_notification(PUSH_TOKEN.to_string(), EXAMPLE_MESSAGE.to_string());
```
heilhead, line 35: Maybe this should now be an async test to be able to await `send_notification()` result?

I don't think that `send_notification()` would event execute without an `await`.

heilhead (approved):
Looks good! Left a couple minor comments for possible future refactoring.

HarryET:
Thanks @heilhead for the fixes, and the refactors have been places into #5 to be done at a later date.

Cargo.toml
HarryET: There were docs for it and it's in `rc` so wanted to save future refactors once it becomes stable.

PR #10, opened 2022-09-24
feat: Postgres

HarryET:
# Description
Adds Postgres as the primary datastore

## How Has This Been Tested?
Locally tested

## Due Diligence

* [x] Breaking change


arein (dismissed):
Readme instructions on how to run the db locally would be great or even better a docker compose 

Maybe a follow up PR

src/store.rs
```diff
@@ -1,35 +1,52 @@
...
+        self.execute(query).await?;
+
         Ok(())
     }
 
-    fn get_client(&self, id: &str) -> error::Result<Option<&Client>> {
-        let client = self.get(&K::from(id.to_string()));
-        Ok(client)
+    async fn get_client(&self, _id: &str) -> error::Result<Option<Client>> {
+        // TODO Get Clients
```
heilhead: Shouldn't this be implemented before merging?

heilhead (approved):
Looks good except the unimplemented part of the SQL storage. Should we address that before merging, what's the plan?

heilhead (dismissed):

docker-compose.yml
```diff
@@ -3,5 +3,20 @@ services:
...
+  echo-server:
+      build: .
+      ports:
+        - "3000:3000"
+      environment:
+        - PORT=3000
+        - LOG_LEVEL=INFO
+        - DATABASE_URL=postgres://postgres:root@postgres:5432/postgres
+        - TELEMETRY_ENABLED=true
+        - TELEMETRY_GRPC_URL=http://jaeger:4317
```
arein, line 22: neat

src/main.rs
```diff
@@ -46,7 +36,20 @@ async fn main() -> error::Result<()> {
...
+        .expect("database url is missing");
+
+    let store = PgPoolOptions::new()
+        .max_connections(5)
+        .connect(database_url)
+        .await?;
+
+    // Run database migrations. `./migrations` is the path to migrations, relative to the root dir (the directory
+    // containing `Cargo.toml`).
+    sqlx::migrate!("./migrations").run(&store).await?;
```
arein, line 51: what happens if we run multiple instances of the app?

src/providers/noop.rs
```diff
@@ -41,12 +39,12 @@ impl NoopProvider {
...
-#[cfg(test)]
-impl NoopProvider {
-    /// Get notifications for a specific token
-    pub fn get_notifications(&mut self, token: String) -> &Vec<String> {
-        self.bootstrap(token.clone());
-
-        self.notifications.get(&token).unwrap()
-    }
-}
+// #[cfg(test)]
```
arein, line 42: obvio no dead code - cool to fix later

arein (approved):

src/main.rs
HarryET: It has a migrations table, it will ignore as long as they are in-sync

src/providers/noop.rs
HarryET: Yeah, I'll add it too the refactor issue

docker-compose.yml
HarryET: :)

src/main.rs
heilhead: > what happens if we run multiple instances of the app?

There's a [locking mechanism](https://github.com/launchbadge/sqlx/blob/main/sqlx-core/src/migrate/migrator.rs#L70-L81) (enabled by default), which should prevent race conditions.

PR #19, opened 2022-09-30
feat: Webhook Validation

HarryET:
# Description
Validates webhooks from the relay so that only valid webhooks are processed

## How Has This Been Tested?
N/a needs integrations tests (future PR)

## Due Dilligence

* [x] Breaking change


HarryET:
Struggling with some of the request issues with the middleware, will continue looking tomorrow.

src/main.rs
```diff
@@ -52,6 +55,12 @@ async fn main() -> error::Result<()> {
 
     let mut state = state::new_state(config, store)?;
 
+    // Fetch public key so it's cached for the first 6hrs
+    let public_key = state.relay_client.public_key().await;
+    if public_key.is_err() {
+        warn!("Failed initial fetch of Relay's Public Key, this may prevent webhook validation.")
+    }
```
arein, line 62: so echo server will still boot if relay is down? (correct behavior as relay is an upstream not downstream dependency)

arein (dismissed):

Cargo.toml
```diff
@@ -10,28 +10,42 @@ build = "build.rs"
 [dependencies]
 tokio = { version = "1", features = ["full"] }
 axum = { version = "0.6.0-rc.2", features = ["json"] }
+hyper = "0.14.20"
```
heilhead: It's best to not specify the 'patch' part of the version, e.g. `0.14` in this case. This applies to all dependencies, since it makes it easier for cargo to resolve compatible versions in the dependency tree and minimize duplication.

src/middleware/validate_signature.rs
```diff
@@ -0,0 +1,147 @@
...
+                description: "Failed to parse signature from bytes".to_string(),
+                location: ErrorLocation::Header
+            }]))),
+        ));
+    }
+
+    Ok(matches!(
+        public_key.verify(sig_body.as_bytes(), &sig.unwrap()),
+        Ok(_)
+    ))
```
heilhead:
```suggestion
    Ok(public_key.verify(sig_body.as_bytes(), &sig.unwrap()).is_ok())
```

src/state.rs
```diff
@@ -1,16 +1,32 @@
...
+use std::sync::Arc;
 use tracing_subscriber::prelude::*;
 
+#[derive(Clone)]
 pub struct Metrics {
     pub registered_webhooks: UpDownCounter<i64>,
     pub received_notifications: Counter<u64>,
 }
 
+pub trait State<S>
```
heilhead, line 17: Do we really need this trait? I feel like axum should have a simpler way of passing the state down to middleware, but I'm not aware of it...

src/state.rs
```diff
@@ -1,16 +1,32 @@
...
+where
+    S: ClientStore,
+{
+    fn get_config(&self) -> Config;
```
heilhead: The naming convention is to omit `get_` prefix for getters, and have only the `set_` prefix for setters.

src/state.rs
```diff
@@ -40,6 +57,7 @@ where
         store,
         providers,
         supported_providers,
+        relay_client: RelayClient::new("https://relay.walletconnect.com".to_string()),
```
heilhead, line 60: Should the URL not be an env var to support multiple/custom envs?

/cc @arein

heilhead (dismissed):
Looks good! Left a couple of bikeshedding comments.

src/main.rs
HarryET: Yes, if the relay is down it will boot but warn it couldn't get the key. However webhooks can't be verified and therefore can't be delivered which is fine as the relay would do that anyway.

Cargo.toml
HarryET: Resolved

src/state.rs
HarryET: There is a planned config refactor, saving this comment for that

src/state.rs
HarryET: Because we aren't makeing it as a middleware and as an extractor we can only add extra requirements to the `S` state type so I made the trait so we are using the intended was of passing state

PR #23, opened 2022-10-11
feat: move clippy config to config file and stop using in CI

### LINKED ISSUE #9: enhancement: Clippy Config
Create a clippy config and stop using `clippy::all` in CI

thaodt:
# Description

move clippy checks to config file and stop using in CI.

Resolves #9 

## How Has This Been Tested?

Run `cargo clippy` in your local machine and see result.

## Due Diligence

* [ ] Breaking change

# Suggestion
I think we should still make use of clippy check in CI by add more arguments `-- -D warnings`, this is more strictly. It will notify us for each PR is created. Warnings are fine. It won't be required to pass CI, but it would be great if they are eventually resolved one way or the other. So,why don't we keep this check in CI?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.'cfg(feature = "cargo-clippy")']
+rustflags = [
+  "-Dclippy::all",
+#   "-Dclippy::await_holding_lock"
+]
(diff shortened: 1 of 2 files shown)


HarryET:
Hi. The issue #9 refers to having a more customised Clippy config and that is something we are discussing internally so this PR does not help the project sadly. I have closed it for now and will add that context to the issue.

thaodt:
> Hi. The issue #9 refers to having a more customised Clippy config and that is something we are discussing internally so this PR does not help the project sadly. I have closed it for now and will add that context to the issue.

Hi @HarryET okay. Please add more details, im happy to rework this PR.

PR #24, opened 2022-10-11
fix: `database_url` is optional in env

### LINKED ISSUE #18: fix: `database_url` is optional in env
Make `database_url` a required env var

thaodt:
# Description

make `database_url` is required in env.

Resolves #18 

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -23,7 +23,6 @@ pub struct Config {
     pub port: u16,
     #[serde(default = "default_log_level")]
     pub log_level: String,
-    #[serde(default = "default_database_url")]
     pub database_url: String,
     pub telemetry_enabled: Option<bool>,
     pub telemetry_grpc_url: Option<String>,
@@ -63,10 +62,6 @@ fn default_port() -> u16 {
 
 fn default_log_level() -> String {
     "WARN".to_string()
-}
-
-fn default_database_url() -> String {
-    "".to_string()
 }
 
 pub fn get_config() -> error::Result<Config> {
--- a/src/env.rs
+++ b/src/env.rs
@@ -23,7 +23,7 @@ pub struct Config {
     pub port: u16,
     #[serde(default = "default_log_level")]
     pub log_level: String,
-    pub database_url: Option<String>,
+    pub database_url: String,
     pub telemetry_enabled: Option<bool>,
     pub telemetry_grpc_url: Option<String>,
     pub apns_certificate: Option<ApnsCertificateConfig>,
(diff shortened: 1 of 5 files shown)


src/env.rs
```diff
@@ -36,6 +36,10 @@ impl Config {
         tracing::Level::from_str(self.log_level.as_str()).expect("Invalid log level")
     }
 
+    pub fn database_url(&self) -> &str {
```
HarryET: Not needed.

src/main.rs
```diff
@@ -39,14 +39,9 @@ async fn main() -> error::Result<()> {
...
 
-    let database_url = config
-        .database_url
-        .as_deref()
-        .expect("database url is missing");
-
     let store = PgPoolOptions::new()
         .max_connections(5)
-        .connect(database_url)
+        .connect(config.database_url())
```
HarryET: Access config instead

HarryET (changes requested):

src/env.rs
thaodt: removed.

src/main.rs
thaodt: > Access config instead

updated.

src/env.rs
```diff
@@ -64,6 +65,10 @@ fn default_log_level() -> String {
     "WARN".to_string()
 }
 
+fn default_database_url() -> String {
```
HarryET: We don't need a default, if its not there it can panic on startup

HarryET, line 68: We don't need a default, if its not there it can panic on startup

HarryET (changes requested):

src/env.rs
thaodt: @HarryET hmm this change served for `tests`, have you rerun the tests on my PR? Do you want me provide a dummy string instead? If i don't set it, tests will fail.

thaodt:
hi @HarryET , already ping you in conversation above but maybe you didn't notice. Therefore, will add a comment here for details.
This latest commit required changes in `tests` to make it passed. Please double-check.

HarryET:
> hi @HarryET , already ping you in conversation above but maybe you didn't notice. Therefore, will add a comment here for details. This latest commit required changes in `tests` to make it passed. Please double-check.

Hi @thaodt, sorry for the delay been busy. Looks good will have a proper look soon and try merge it!

src/env.rs
HarryET: Yeah I know tests will fail, just update the tests

HarryET (approved):

PR #32, opened 2022-10-20
fix: Only release on manual trigger

HarryET:
# Description
Each push to main was going to trigger a release where-as we only want to release every so often as it rebuilds/redeploys as well as creating new containers

## How Has This Been Tested?
N/a, cannot test actions

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,32 +1,14 @@
 name: release
 
 on:
-  push:
-    branches:
-      - main
-    paths-ignore:
-      - '.github/**'
-      - 'terraform/**'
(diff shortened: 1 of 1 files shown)


arein (approved):
We still need CI/CD on each release. I'd leave as is?

HarryET:
> We still need CI/CD on each release. I'd leave as is?

This just stops it making a release each time, I'm going to do a follow-up PR do we run `clippy`, `fmt` and `test` on each PR but I don't want to re-deploy because it will also publish images which may be depended on by others and it wold be nice to release with multiple features rather than each PR merge being a release.

PR #35, opened 2022-10-21
feat: allow updating device token

### LINKED ISSUE #29: feat: registering clients is overly strict

### Current Behavior

When registering a client that already exists the response is an error

### Requested Behavior

Just respond with a HTTP 200 code

This makes it easier for calling clients

### Anything else?

_No response_

arein:
# Description

This is especially important for iOS: the client id remains in the keychain when the app is uninstalled/reinstalled. Hence, right now the iOS development flow is really hard.

Also: generally a 200 is a much nicer development experience compared to having to handle both a 200 and a 400.

We can handle security things e.g. who can update the token later see #57 

Resolves #29 

## How Has This Been Tested?

Covered in functional tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/register_client.rs
+++ b/src/handlers/register_client.rs
@@ -1,11 +1,7 @@
 use {
     crate::{
         error::{
-            Error::{
-                EmptyField,
-                IncludedTenantIdWhenNotNeeded,
-                ProviderNotAvailable,
-            },
+            Error::{EmptyField, IncludedTenantIdWhenNotNeeded, ProviderNotAvailable},
             Result,
         },
         handlers::Response,
--- a/src/handlers/register_client.rs
+++ b/src/handlers/register_client.rs
@@ -2,7 +2,6 @@ use {
     crate::{
         error::{
             Error::{
-                ClientAlreadyRegistered,
                 EmptyField,
                 IncludedTenantIdWhenNotNeeded,
                 ProviderNotAvailable,
@@ -61,7 +60,15 @@ pub async fn handler(
     };
 
     if exists {
-        return Err(ClientAlreadyRegistered);
+        state
+        .client_store
+        .update_client(&tenant_id, &body.client_id, Client {
+            push_type,
+            token: body.token,
+        })
+        .await?;
+
+        return Ok(Response::default());
     }
 
     state
--- a/src/handlers/register_client.rs
+++ b/src/handlers/register_client.rs
@@ -10,7 +10,7 @@ use {
         },
         handlers::Response,
         state::{AppState, State},
-        stores::{client::Client, StoreError},
+        stores::client::Client,
     },
     axum::extract::{Json, Path, State as StateExtractor},
     serde::{Deserialize, Serialize},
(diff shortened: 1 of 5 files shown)


src/handlers/register_client.rs
```diff
@@ -67,15 +67,9 @@ pub async fn handler(
         );
     }
 
+    // TODO: ideally we allow for clients to update their device tokens
```
HarryET: I'd rather leave this as it is and then have this (or a different) PR change this to upsert functionality as seen in the notifications store

HarryET (changes requested):

src/handlers/register_client.rs
arein: sure

HarryET:
Please address the TODO and resolve the conflicts

arein:
> Please address the TODO and resolve the conflicts

I missed the todo but I think it's to use the DB and to allow actual updates which is what I've implemented.

src/handlers/register_client.rs
```diff
@@ -61,7 +60,15 @@ pub async fn handler(
     };
 
     if exists {
-        return Err(ClientAlreadyRegistered);
+        state
+        .client_store
+        .update_client(&tenant_id, &body.client_id, Client {
```
xav: Isn't there a risk of somebody spoofing the client id and taking over an account?

xav (approved):

src/handlers/register_client.rs
arein: Yes see #57

src/stores/client.rs
```diff
@@ -17,6 +17,7 @@ pub struct Client {
 #[async_trait]
 pub trait ClientStore {
     async fn create_client(&self, tenant_id: &str, id: &str, client: Client) -> stores::Result<()>;
+    async fn update_client(&self, tenant_id: &str, id: &str, client: Client) -> stores::Result<()>;
```
HarryET: Change `create_client` to be an upsert, see notification store

src/handlers/register_client.rs
```diff
@@ -61,7 +60,15 @@ pub async fn handler(
     };
 
     if exists {
```
HarryET: See store comment, then remove this and the accompanying error

HarryET (changes requested):

src/stores/client.rs
```diff
@@ -36,6 +36,7 @@ impl ClientStore for sqlx::PgPool {
                     .push_bind(client.3);
             },
         );
+        query_builder.push(" ON CONFLICT (id) DO UPDATE SET device_token = EXCLUDED.device_token");
```
HarryET, line 39: ~~please also add the ability to update `push_type` on conflict~~ make an issue to update this and then we can merge

HarryET (changes requested):

src/stores/client.rs
arein: I don't see the use case for that but sure

HarryET:
Please run formatter too @arein, `cargo +nightly fmt`

PR #40, opened 2022-11-03
feat: Initial Multi-Tenant Work

HarryET:
# Description
Makes Echo Server multi-tenant and fixes APNS/FCM issues

## How Has This Been Tested?
`dev.echo.walletconnect.com`

## Due Diligence

* [x] Breaking change


arein (approved):
LGTM
What exactly are those APNS fixes?

src/state.rs
```diff
@@ -8,53 +8,60 @@ use opentelemetry::metrics::{Counter, UpDownCounter};
...
     fn supported_providers(&self) -> Vec<ProviderKind>;
     fn relay_client(&self) -> RelayClient;
 }
 
 #[derive(Clone)]
-pub struct AppState<C, N>
-where
-    C: ClientStore,
-    N: NotificationStore,
+pub struct AppState<C, N, T>
```
heilhead: We should refactor this to remove type parameters, and instead use `client_store: Arc<dyn ClientStore>` etc.

HarryET: Refactored!

HarryET:
> LGTM What exactly are those APNS fixes?

This is still WIP, the fixes will be added before merging

arein (approved):
Looks good! Let's come up with a plan of how we write tests for it

HarryET:
> Looks good! Let's come up with a plan of how we write tests for it

Yeah I think there is an issue to re-work the tests!

src/main.rs
```diff
@@ -39,11 +40,19 @@ async fn main() -> error::Result<()> {
     let config =
         env::get_config().expect("Failed to load config, please ensure all env vars are defined.");
 
-    let supported_providers = config.supported_providers();
-    if supported_providers.is_empty() {
-        panic!("You must enable at least one provider.");
+    let mut supported_providers_string = "multi-tenant".to_string();
```
heilhead, line 43: Another way:
```rust
let supported_providers_string = if config.tenant_database_url.is_none() {
    // ...
} else {
    "multi-tenant".to_string()
};
```

src/stores/tenant.rs
```diff
@@ -0,0 +1,139 @@
...
+            Err(e) => Err(e.into()),
+            Ok(row) => Ok(row),
+        }
+    }
+}
+
+pub struct DefaultTenantStore(Tenant);
+
+impl DefaultTenantStore {
+    pub fn new(config: Arc<Config>) -> Result<DefaultTenantStore> {
```
heilhead, line 120: Why does it need an `Arc<Config>` if it only clones a bunch of values out of `Config`? `&Config` here would be enough.

src/main.rs
```diff
@@ -58,7 +67,34 @@ async fn main() -> error::Result<()> {
     // containing `Cargo.toml`).
     sqlx::migrate!("./migrations").run(&store).await?;
 
-    let mut state = state::new_state(config, store.clone(), store.clone())?;
+    let mut tenant_store: TenantStoreArc =
```
heilhead, line 70: Same here:
```rust
let tenant_store = if let Some(tenant_database_url) = &config.tenant_database_url {
    // ...
} else {
    // ...
};
```

src/main.rs
```diff
@@ -58,7 +67,34 @@ async fn main() -> error::Result<()> {
...
+        Arc::new(DefaultTenantStore::new(Arc::new(config.clone()))?);
```
heilhead, line 71: As noted below, it doesn't really need an `Arc<Config>`, and could use `&Config` here.

src/state.rs
```diff
@@ -1,82 +1,72 @@
...
 use crate::stores::notification::NotificationStore;
-use crate::{env::Config, providers::ProviderKind};
+use crate::stores::tenant::TenantStore;
 use build_info::BuildInfo;
 use opentelemetry::metrics::{Counter, UpDownCounter};
 use opentelemetry::sdk::trace::Tracer;
 use std::sync::Arc;
 use tracing_subscriber::prelude::*;
 
+pub type ClientStoreArc = Arc<dyn ClientStore + Send + Sync + 'static>;
```
heilhead, line 12: `Send + Sync + 'static` can be supertraits of `ClientStore`, to simplify these types, e.g.:
```rust
trait ClientStore: Send + Sync + 'static {
    // ...
}
```

src/state.rs
```diff
@@ -1,82 +1,72 @@
...
-    N: NotificationStore,
-{
+pub trait State {
     fn config(&self) -> Config;
     fn build_info(&self) -> BuildInfo;
-    fn client_store(&self) -> C;
-    fn notification_store(&self) -> N;
-    fn providers(&self) -> Providers;
-    fn supported_providers(&self) -> Vec<ProviderKind>;
+    fn client_store(&self) -> ClientStoreArc;
```
heilhead, line 25: A more ergonomic API would be to return `&Arc<dyn ClientStore>` and leave the decision of whether to clone it up to the consumer. Or even `&dyn ClientStore` if the consumers weren't going to clone it.

src/stores/tenant.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    pub fn provider(&self, provider: &ProviderKind) -> Result<Provider> {
+        if !self.providers().contains(provider) {
+            return Err(ProviderNotAvailable(provider.into()));
+        }
+
+        match provider {
+            ProviderKind::Apns => {
+                let endpoint = match self.apns_sandbox {
```
heilhead, line 56: Why not `if ... else`?

src/stores/tenant.rs
```diff
@@ -0,0 +1,139 @@
...
+                            endpoint,
+                            topic.clone(),
+                        )?;
+
+                        Ok(Apns(apns_client))
+                    }
+                    _ => Err(ProviderNotAvailable(provider.into())),
+                }
+            }
+            ProviderKind::Fcm => match self.fcm_api_key.clone() {
```
heilhead, line 81: A cleaner approach would be:
```rust
self.fcm_api_key.clone()
    .map(|api_key| Fcm(FcmProvider::new(api_key)))
    .ok_or_else(|| ProviderNotAvailable(provider.into()))
```

Issue #13, opened 2022-09-26
enhancement: Integration Tests


arein:
can be in the sample apps or better a small integration test suite here in the repository.

Personally I like the mocha for integ tests e.g. https://github.com/WalletConnect/registry-api-cf-workers/tree/master/integration

thaodt:
> can be in the sample apps or better a small integration test suite here in the repository.
> 
> Personally I like the mocha for integ tests e.g. https://github.com/WalletConnect/registry-api-cf-workers/tree/master/integration

@arein the ref link seems not found now. Could you please update? how about using [bats](https://github.com/bats-core/bats-core) for integration tests?

arein:
@HarryET any ideas how to actually test this? I think it needs to be e2e with the SDKs

HarryET:
Yeah E2E, or we can have a local suite which calls all the endpoints but maybe that can be part of unit tests

PR #49, opened 2022-11-17
feat: refactor for functional test

arein:
# Description

In order to spin up a server in the functional test we need to decouple the server logic from the `main` logic. This PR does that.

Towards #13 

## How Has This Been Tested?

`cargo test`

## Due Diligence

* [ ] Breaking change


src/lib.rs
```diff
@@ -6,3 +29,172 @@ pub mod providers;
 pub mod relay;
 pub mod state;
 pub mod stores;
+
+pub async fn bootstap(
+    mut state: AppState,
+    supported_providers_string: String,
```
HarryET: Should be figured out from the AppState

src/main.rs
```diff
@@ -1,44 +1,19 @@
...
 #[tokio::main]
-async fn main() -> error::Result<()> {
+async fn main() -> Result<()> {
     dotenv().ok();
-    let config =
-        env::get_config().expect("Failed to load config, please ensure all env vars are defined.");
+    let config = echo_server::env::get_config()
+        .expect("Failed to load config, please ensure all env vars are defined.");
 
     let mut supported_providers_string = "multi-tenant".to_string();
```
HarryET: Move inside of bootstrap function

src/main.rs
```diff
@@ -1,44 +1,19 @@
...
-use tracing::{warn, Level};
-use tracing_subscriber::fmt::format::FmtSpan;
 
 #[tokio::main]
-async fn main() -> error::Result<()> {
+async fn main() -> Result<()> {
     dotenv().ok();
-    let config =
-        env::get_config().expect("Failed to load config, please ensure all env vars are defined.");
+    let config = echo_server::env::get_config()
```
HarryET: Why is this now scoped?

src/main.rs
```diff
@@ -89,174 +64,12 @@ async fn main() -> error::Result<()> {
         tenant_store = Arc::new(tenant_database);
     }
 
-    let mut state = state::new_state(
+    let state = echo_server::state::new_state(
```
HarryET: Why is this now scoped?

src/main.rs
```diff
@@ -89,174 +64,12 @@ async fn main() -> error::Result<()> {
...
-    println!("{}", header);
-
-    let addr = SocketAddr::from(([0, 0, 0, 0], port));
-    axum::Server::bind(&addr)
-        .serve(app.into_make_service())
-        .await
-        .unwrap();
-
-    Ok(())
+    echo_server::bootstap(state, supported_providers_string).await
```
HarryET: Please include a `?` so that the error is bubbled up

HarryET (changes requested):
Looks good, a few comments and questions though

HarryET:
Please also run Clippy on the PR and see if Clippy is upset with any of your changes. If it is please fix, else we can have a follow-up PR

src/main.rs
arein: No I need to be able to not use `env` when bootstrapping from functional test

src/main.rs
arein: Was following compiler recommendations and don't see a problem with it?

src/main.rs
HarryET: `env::get_config()` just provides a `Config` which is what this string is generated from. If you are mocking a `Config` then it will mock this variable too.

src/main.rs
HarryET: No problems, just wondered if you had a specific reason why :)

src/lib.rs
```diff
@@ -6,3 +33,230 @@ pub mod providers;
 pub mod relay;
 pub mod state;
 pub mod stores;
+
+pub async fn bootstap() -> error::Result<()> {
```
HarryET: This should take a `config: Config` param, no? That way you don't need any env vars when you bootstrap for tests?

src/main.rs
```diff
@@ -1,262 +1,4 @@
...
-
-    let addr = SocketAddr::from(([0, 0, 0, 0], port));
-    axum::Server::bind(&addr)
-        .serve(app.into_make_service())
-        .await
-        .unwrap();
-
-    Ok(())
+async fn main() -> echo_server::error::Result<()> {
+    Ok(echo_server::bootstap().await?)
```
HarryET:
```suggestion
    dotenv().ok();
    let config = env::get_config()
        .expect("Failed to load config, please ensure all env vars are defined.");
    Ok(echo_server::bootstap(config).await?)
```

Pass config into `bootstrap(...)` so that in tests you can specify the config

HarryET (changes requested):

HarryET (approved):
LGTM, thanks for addressing feedback

PR #53, opened 2022-11-21
feat: cover registration logic

arein:
# Description

Covers registration logic in functional test.

Towards #13 

## How Has This Been Tested?

`cargo test`

## Due Diligence

* [ ] Breaking change


src/lib.rs
```diff
@@ -169,7 +169,7 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
                 received_notifications: notification_counter,
             },
         )
-    } else {
+    } else if state.config.environment != "test" {
```
HarryET: not needed please use `cfg!`

src/stores/client.rs
```diff
@@ -60,15 +60,17 @@ impl ClientStore for sqlx::PgPool {
...
             sqlx::QueryBuilder::new("DELETE FROM public.notifications WHERE client_id = ");
         notification_query_builder.push_bind(id);
-        notification_query_builder.push("and tenant_id = ");
+        notification_query_builder.push(" and tenant_id = ");
         notification_query_builder.push_bind(tenant_id);
         let notification_query = notification_query_builder.build();
 
         self.execute(notification_query).await?;
 
+        println!("Deleting client {}", id);
```
HarryET: Please remove debug lines or use `tracing`

src/stores/client.rs
```diff
@@ -60,15 +60,17 @@ impl ClientStore for sqlx::PgPool {
...
+
         let mut query_builder = sqlx::QueryBuilder::new("DELETE FROM public.clients WHERE id = ");
         query_builder.push_bind(id);
-        query_builder.push("and tenant_id = ");
+        query_builder.push(" and tenant_id = ");
```
HarryET, line 74: Why spaces?

src/stores/client.rs
```diff
@@ -60,15 +60,17 @@ impl ClientStore for sqlx::PgPool {
         let mut notification_query_builder =
             sqlx::QueryBuilder::new("DELETE FROM public.notifications WHERE client_id = ");
         notification_query_builder.push_bind(id);
-        notification_query_builder.push("and tenant_id = ");
+        notification_query_builder.push(" and tenant_id = ");
```
HarryET, line 66: why spaces?

HarryET (changes requested):

src/stores/client.rs
arein: ups. thank you!

src/stores/client.rs
arein: did this work before? It failed unless there was a space for me.

src/stores/client.rs
arein: see above

src/lib.rs
arein: How?... 🙏

src/stores/client.rs
HarryET: Should work, but happy to merge with this and investigate later 😄

src/lib.rs
HarryET: See `./providers/mod.rs` for examples using the `#[cfg()]` or here are some doc links: https://doc.rust-lang.org/rust-by-example/attribute/cfg.html, https://doc.rust-lang.org/reference/conditional-compilation.html

src/lib.rs
```diff
@@ -27,6 +27,8 @@ use crate::{
     state::{Metrics, TenantStoreArc},
     stores::tenant::DefaultTenantStore,
 };
+
+#[cfg(not(any(debug_assertions, test)))]
```
HarryET: why?

src/lib.rs
```diff
@@ -170,6 +172,7 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
             },
         )
     } else {
+        #[cfg(not(any(debug_assertions, test)))]
```
HarryET: why?

HarryET (changes requested):

src/lib.rs
arein: We cannot have multiple instances of the server running with logging. It barfs about logging context if I try to do so.

Want me to add a comment?

src/lib.rs
arein: Same as above

HarryET: Why did you not just use `else if cfg!() { /* ... */ }`

src/lib.rs
HarryET: This disables them even just in debugging mode. Also see comment below

arein: I assumed this only disables when in `test`. Is there a way to do that?

src/lib.rs
arein: I didn't know about that syntax.

src/error.rs
```diff
@@ -88,6 +88,12 @@ pub enum Error {
 
     #[error("invalid tenant id: {0}")]
     InvalidTenantId(String),
+
+    #[error("server didn't boot within timeout")]
+    ServerBootTimeout(String),
+
+    #[error("server didn't shut down within timeout")]
+    ServerShutdownTimeout(String),
```
HarryET: Thease should be removed. They aren't used.

src/lib.rs
```diff
@@ -169,7 +171,8 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
                 received_notifications: notification_counter,
             },
         )
-    } else {
+    } else if cfg!(not(test)) {
+        #[cfg(not(any(debug_assertions, test)))]
```
HarryET: This can be removed now that you have the `if cfg!()` syntax above

HarryET (changes requested):

src/env.rs
```diff
@@ -16,6 +16,8 @@ pub struct Config {
     pub tenant_database_url: Option<String>,
     #[serde(default = "default_tenant_id")]
     pub default_tenant_id: String,
+    #[serde(default = "default_is_test")]
+    pub is_test: bool,
```
HarryET, line 21: Please use serde ignore so that people cannot send this as an env var and hide it at the bottom of the config struct

HarryET (changes requested):

src/env.rs
arein: You mean `serde skip`? https://serde.rs/field-attrs.html#skip

HarryET: yah

src/env.rs
```diff
@@ -16,6 +16,8 @@ pub struct Config {
     pub tenant_database_url: Option<String>,
     #[serde(default = "default_tenant_id")]
     pub default_tenant_id: String,
+    #[serde(default = "default_is_test", skip)]
+    pub is_test: bool,
```
HarryET, line 21:
```suggestion
    /// This is an internal flag to disable logging, cannot be defined by user
    pub is_test: bool,
```

HarryET (changes requested):

PR #62, opened 2022-12-08
feat: add validate workflow

arein:
# Description

Runs integration tests in CD.

Towards #13 

## How Has This Been Tested?

Rand tests

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -125,7 +125,8 @@ jobs:
     uses: ./.github/workflows/validate.yml
     with:
       environment: 'staging'
-    secrets: inherit
+      secrets:
+        TEST_TENANT_ID_APNS: ${{ secrets.TEST_TENANT_ID_APNS }}
 
   deploy-infra-prod:
     runs-on: ubuntu-latest
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -120,15 +120,12 @@ jobs:
           onepassword-api-key: ${{ secrets.OP_CONNECT_TOKEN }}
           app-name: ${{ github.event.repository.name }}
 
-  # TODO: Run integration test suite here
-
-  check-staging-health:
-    needs:
-      - deploy-infra-staging
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check Staging Up
-        run: curl --fail --connect-timeout 15 --retry 5 --retry-delay 5 https://staging.echo.walletconnect.com/health
+  validate_staging:
+    needs: [deploy-infra-staging]
+    uses: ./.github/workflows/validate.yml
+    with:
+      environment: 'staging'
+    secrets: inherit
 
   deploy-infra-prod:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


xav (approved):

integration/integration.test.ts
```diff
@@ -2,16 +2,18 @@ import axios from 'axios'
...
 }
 
 const BASE_URLS = new Map<string, string>([
+  ['prod', 'https://echo.walletconnect.com'],
   ['staging', 'https://staging.echo.walletconnect.com'],
   ['dev', 'http://localhost:3000'],
 ])
 
-const TEST_TENANT = 'f1e0fcb9-75f8-49b5-a05b-00c35ac8418f'
+const TEST_TENANT = process.env.TEST_TENANT_ID_APNS
```
HarryET, line 16: Can we make an issue to test FCM as well as APNS?

.github/workflows/cd.yml
```diff
@@ -120,15 +120,13 @@ jobs:
...
-    steps:
-      - name: Check Staging Up
-        run: curl --fail --connect-timeout 15 --retry 5 --retry-delay 5 https://staging.echo.walletconnect.com/health
+  validate_staging:
+    needs: [deploy-infra-staging]
+    uses: ./.github/workflows/validate.yml
+    with:
+      environment: 'staging'
+      secrets:
+        TEST_TENANT_ID_APNS: ${{ secrets.TEST_TENANT_ID_APNS }}
```
HarryET, line 129: Have you added it as a secret?

.github/workflows/cd.yml
```diff
@@ -120,15 +120,13 @@ jobs:
           onepassword-api-key: ${{ secrets.OP_CONNECT_TOKEN }}
           app-name: ${{ github.event.repository.name }}
 
-  # TODO: Run integration test suite here
-
-  check-staging-health:
```
HarryET, line 125: we should keep this step but edit it to fetch staging health and check for version to make sure we don't run tests till the version matches the new one else we could be running tests on an old container

HarryET (changes requested):

HarryET:
Also can you change the environment name in `ci_terraform.yml` to just `staging` so that we don't have 2 staging environments? Or can we remove an environment from the terraform ci as we don't really need one? We can also just make an issue to do this later.

.github/workflows/cd.yml
arein: Yes.

integration/integration.test.ts
arein: Yup it was my plan to do as a fast follow-up. https://github.com/WalletConnect/echo-server/issues/63

.github/workflows/cd.yml
arein: CD only deploys via Terraform. I fixed the Terraform to [`wait_for_completion`](https://github.com/WalletConnect/echo-server/blob/main/terraform/ecs/main.tf#L112) so now we know that by the time the deploy step is done the latest version is deployed.
So this is not an issue.

arein:
> Also can you change the environment name in `ci_terraform.yml` to just `staging` so that we don't have 2 staging environments? Or can we remove an environment from the terraform ci as we don't really need one? We can also just make an issue to do this later.

Out of scope for this PR but sure I've updated it.

.github/workflows/cd.yml
HarryET: 🤝

HarryET (approved):

PR #67, opened 2022-12-19
feat: add basic dashboard

arein:
# Description

Adds an empty dashboard we can then add things onto.

## How Has This Been Tested?

Applied to staging.

## Due Diligence

* [ ] Breaking change


terraform/monitoring/main.tf
```diff
@@ -18,4 +18,54 @@ resource "grafana_data_source" "cloudwatch" {
   json_data {
     default_region = "eu-central-1"
   }
-}
\ No newline at end of file
+}
+
+resource "grafana_dashboard" "at_a_glance" {
```
HarryET, line 23: Nitpick: Could a better name be used here or is this standard across all repos?

HarryET (changes requested):

terraform/monitoring/main.tf
arein: This name is not visible to end-users in Grafana. I've used this everywhere thus far but no need to keep it like that. Also don't find it confusing to developers who are the only ones seeing this when editing monitoring so also don't think it's important to find a better name.

HarryET: Ah thats fine then, thought it was potentially visible

HarryET (approved):

PR #70, opened 2022-12-20
feat: Migrate to E2EE webhooks

HarryET:
# Description
Changes the spec to support E2EE payloads, based on work done in push meeting yesterday and after a meeting with @Rakowskiii this morning. This unblocks #68 here as well as the work on the relay.

## How Has This Been Tested?
N/a, spec changes only

## Due Diligence

* [x] Breaking change


Rakowskiii (approved):

spec/spec.md
```diff
@@ -80,21 +80,24 @@ DELETE <ECHO_SERVER_URL>/clients/<CLIENT_ID>
...
+- `payload`: A string with 3 values encoded and split with a `:` as a delimiter.
+  - `topic`: This is used by the SDKs to decrypt encrypted payloads on the client side
+  - `flags`: an integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
+    ```
+    Encrypted  = 1 << 0
+    Sign       = 1 << 1
+    Auth       = 1 << 2
+    Chat       = 1 << 3
+    Push       = 1 << 4
+    ```
```
llbartekll, line 105: don't get this part. Could it be just a binary value? 0 - unencrypted, 1 - encrypted ?

spec/spec.md
```diff
@@ -80,21 +80,24 @@ DELETE <ECHO_SERVER_URL>/clients/<CLIENT_ID>
...
-
 ```
 POST <ECHO_SERVER_URL>/clients/<CLIENT_ID>
 {
     "id": "0000-0000-0000-0000"
-    "payload": {
-        "title": "New Request",
-        "description": "You have a signing request."
-    }
+    "payload": "<topic>:<flags>:<blob>"
```
arein: this now requires us to ensure that `topic`/`flags`/`blob` don't contain any `:`.

IMO that's an unnecessary constraint and may make debugging later on kinda hard in case there is a `:` in any of those.

arein (changes requested):

spec/spec.md
HarryET: Would you rather this is an object?

spec/spec.md
HarryET: It could but do we not want to be able to tell the SDKs apart?

spec/spec.md
arein: Yes i think we save us some potential trouble by making it an object.

HarryET: Okay, I'll update to make it an object, it was a string as @Rakowskiii said that was something agreed/discussed

spec/spec.md
```diff
@@ -80,21 +80,27 @@ DELETE <ECHO_SERVER_URL>/clients/<CLIENT_ID>
...
-> This will have E2E encryption in the future but for now is designed without encryption.
-
 ```
 POST <ECHO_SERVER_URL>/clients/<CLIENT_ID>
 {
     "id": "0000-0000-0000-0000"
     "payload": {
-        "title": "New Request",
-        "description": "You have a signing request."
+      "topic": "<topic>",
```
xav, line 88: Should we explain what the topic is (the hash of blah blah blah...)?

spec/spec.md
```diff
@@ -80,21 +80,27 @@ DELETE <ECHO_SERVER_URL>/clients/<CLIENT_ID>
...
 ```
 
 - `CLIENT_ID`: This is used to map to the device token
 - `id`: This is generated by the Relay to prevent duplicate notifications
-- `payload`: A JSON object that holds the relevant payload data. **NOTE:** This will become a string that is encrypted
-  by the dApp and not decrypted until received by the client
\ No newline at end of file
+- `payload`: A string with 3 values encoded and split with a `:` as a delimiter.
+  - `topic`: This is used by the SDKs to decrypt encrypted payloads on the client side
+  - `flags`: an integer value that uses bitmasks to encode information about the payload for the SDKs, the flags are below
```
xav, line 98: Let's make sure we stay within 1 byte if we use bit masks to avoid endianness issues.

xav (approved):

spec/spec.md
llbartekll: From client perspective, I don’t think it’s necessary

arein (approved):

spec/spec.md
HarryET: Thats okay, this is going to be moved into the main specs at some point and we can link it then

spec/spec.md
HarryET: 👍🏻 Might remove it if we decide it's unneeded

spec/spec.md
HarryET: Wallets wanted a way to distinguish them, cc @arein, I believe rainbow wanted to be able to tell if they were getting them from us or from potentially a third-party

llbartekll: note that for chat you can have two types of notifications: 
chat_invite - unencrypted
chat_message - encrypted

how echo server decides if it should set the mutable_content flag (for iOS) if it only knows the api(Chat)?
correct me if I am wrong but cannot see other param.

but so far cannot find a usecase for an end client to need PN related sdk - all encrypted will go to we3inbox

HarryET: It can set 2 flags at once, it uses bit masking so the request can be encrypted and chat, just chat, encrypted and push, just push, etc

llbartekll: unencrypted will just appear on the screen

HarryET: Yes

llbartekll: you can decide, I don't need to know the sdk right now but no one knows if we need it in the future

HarryET: I think we should keep it for now as it gives more information to Echo Server

llbartekll (approved):

Talhaali00 (approved):

PR #71, opened 2022-12-20
fix: use self instead of unnecessary structure names

farazdagi:
# Description

When looking into `echo-server` implementation I've noticed that sometimes structure names are being used where `Self` can be used instead (clippy has it under [use_self](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) lint). So, I've run linter to see all the places that can be updated, hence this PR. 

## How Has This Been Tested?

When running `cargo test` only 3 tests are located, so I'm not sure how this should be tested (I've made sure that code whenever updated makes sense -- not sure if that is enough).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -97,49 +97,49 @@ pub enum Error {
 impl IntoResponse for Error {
     fn into_response(self) -> Response {
         match self {
-            Error::Apns(e) => crate::handlers::Response::new_failure(StatusCode::INTERNAL_SERVER_ERROR, vec![
+            Self::Apns(e) => crate::handlers::Response::new_failure(StatusCode::INTERNAL_SERVER_ERROR, vec![
                 ResponseError {
                     name: "apns".to_string(),
                     message: e.to_string(),
                 }
             ], vec![]),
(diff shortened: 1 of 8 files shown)


HarryET (commented):
This hasn't caused clippy issues before and I'd argue it adds extra clarity

farazdagi:
It is definitely not something that clippy will complain about (it is in `clippy::nursery`).

When it comes to extra clarity, I guess it is a matter of preference.

However, even if it adds clarity, it seems that pattern is not used consistently, for instance, sometimes structure names are used

https://github.com/WalletConnect/echo-server/blob/611f14c04aa1fe94d39e3e0bbce0a4a25249886f/src/relay/mod.rs#L18-L19

and sometimes `Self` is used as return type, and then structure name when actually returning:
https://github.com/WalletConnect/echo-server/blob/611f14c04aa1fe94d39e3e0bbce0a4a25249886f/src/providers/fcm.rs#L15-L16

So, you might want to handle those cases consistently. As for this PR, let's close it w/o merging.

HarryET:
👌🏼 I think we should discuss as a team what is the preference then change all repositories accordingly!

PR #72, opened 2022-12-22
feat: E2EE

HarryET:
# Description
Support E2EE push notifications

## How Has This Been Tested?
New unit tests written in PR

## Due Diligence

* [x] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,10 +4,11 @@ version = 3
 
 [[package]]
 name = "a2"
-version = "0.6.2"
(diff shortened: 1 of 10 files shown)


arein (approved):
Do we need a README update?

Is the use-case of static sign-notifications still supported?

HarryET:
> Do we need a README update?

I'll check, still got some tests to add to this PR

> Is the use-case of static sign-notifications still supported?

Yes but the format has now changed. This is documented in the `spec.md`

PR #78, opened 2023-01-11
feat: Statically compile echo-server and run on scratch image

welps:
# Description

This is an opinionated PR that was finagled to fit the current approach set in the `Dockerfile` so feel free to reject or manipulate into something that works. We prefer to statically compile binaries so we can run them on `scratch`, when possible, to minimize surface area for attacks.

The previous build behavior, though not exact, is possible by supplying `runtime=debian:buster-slim` and `target=''` as build args. 

There are things to think about regarding the `Dockerfile` and its intended usage since there are currently some ideas that are incompatible/awkward:
- `runtime` is a build arg which implies it can be changed, but before this PR, compilation would fail on non-debian based distros due to `apt-get` usage 
     - After this PR, it make less sense since echo-server is compiled statically and we copy all necessary files over from the `build` stage. Providing `runtime=debian:buster-slim` will still work though! 
- The introduction of `target` by me isn't an ideal solution because (providing `target='' will still work though`): 
     - different targets have their own system dependencies that need to be installed in the `build` stage 
     - different `runtime` options may be incompatible with the `target` (e.g., mismatched archs)
    
## How Has This Been Tested?

I brought up services via docker-compose and verified echo-server boots, handles a request correctly, and is terminated properly.

I tested with following combos:
- `target=''` and `runtime=debian:buster-slim`
- `target=x86_64-unknown-linux-musl` and `runtime=scratch`

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -558,7 +558,7 @@ dependencies = [
 
 [[package]]
 name = "echo-server"
-version = "0.11.4"
+version = "0.11.5"
 dependencies = [
  "a2",
  "async-trait",
@@ -574,6 +574,7 @@ dependencies = [
  "futures-util",
  "hex",
  "hyper",
+ "openssl",
  "opentelemetry",
  "opentelemetry-otlp",
  "opentelemetry-prometheus",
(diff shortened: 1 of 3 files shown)


HarryET:
Hey @welps, posted some feedback/response in slack! To sumerise, if you have a better docker file feel free to change this PR to include that - provided it still can create containers to be run in the same way we do now.

HarryET:
Hi @welps, just following up

PR #86, opened 2023-01-29
feat: migrate `bat-cave` to `echo-server`

HarryET:
# Description
Experiment to move https://github.com/walletconnect/bat-cave into Echo Server so that all code is public and there is less to maintain

## How Has This Been Tested?
N/a

## Due Diligence

* [x] Breaking change


arein (commented):
lgtm

arein (approved):

PR #88, opened 2023-01-31
feat: CORS

### LINKED ISSUE #87: feat: CORS headers

### Current Behavior

Some CORS headers are missing which makes it impossible to make requests from browsers.
`Reason: header ‘authorization’ is not allowed according to header ‘Access-Control-Allow-Headers’ from CORS preflight response`

### Requested Behavior

Add CORS headers so that we can make requests from browsers

### Anything else?

See example [here](https://www.stackhawk.com/blog/rust-cors-guide-what-it-is-and-how-to-enable-it/)

HarryET:
# Description
Closes #87 for @Cali93 

> **Note**
> Looks like this broke the rust compiler on nightly, issue open [here](https://github.com/rust-lang/rust/issues/107512)

## How Has This Been Tested?
N/a

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,9 @@ LOG_LEVEL_OTEL=info,echo-server=trace
 TENANT_DATABASE_URL=
 DEFAULT_TENANT_ID= # This has a default value and dosen't hold much impact to the running of echo-server
 
+# CORS
+CORS_ALLOWED_ORIGINS=*
+
 # Telemetry
 # OTLP ENV also supported e.g.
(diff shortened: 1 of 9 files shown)


HarryET:
Right now we allow all origins, @Cali93 do you have a list of origins you would like on prod vs staging? and dev will always be `*`

Cali93:
@HarryET  Yes, here are the domains to allow:

- dev/staging: * 
- prod: https://cloud.walletconnect.com

arein (approved):

PR #95, opened 2023-02-03
fix: lowercase string conversions into ProviderKind

welps:
# Description

Since push notification registration is not handled by the SDK, the implementation burden is on the wallet, and open to input errors like sending `type` as `FCM` and not `fcm`. This is not worth rejecting as a validation error since it is reconcilable by the echo-server backend.

## How Has This Been Tested?

I manually tested this, it's a pretty straightforward fix.

## Due Diligence

* [ ] Breaking change


HarryET (approved):
Thanks Wayne!

PR #96, opened 2023-02-03
fix: Strip out decentralized identifier prefix from client_id

welps:
# Description

https://github.com/WalletConnect/echo-server/blob/main/Contributors.md#client-registration suggests that you can just pass the output of the SDK function that returns Client ID, but this is not true for at least the javascript SDK, which returns the client ID in 
the format `did:key:foo`. The echo-server will happily accept this, but will prove to be problematic when the notification endpoint gets client ID without this prefix.

This PR just strips out this prefix for the three endpoints.

## How Has This Been Tested?

Manually tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/delete_client.rs
+++ b/src/handlers/delete_client.rs
@@ -1,5 +1,10 @@
 use {
-    crate::{error::Result, handlers::Response, log::prelude::*, state::AppState},
+    crate::{
+        error::Result,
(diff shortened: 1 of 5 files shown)


HarryET (changes requested):
Please do the stripping of the prefix in the actual handlers not the single tenant wrapper

HarryET (approved):
Will merge when I'm back on Tuesday

PR #97, opened 2023-02-05
fix(hotfix): dedupe messages

arein:
# Description

Currently messages with the same id aren't deduped even though existing code leads to suggest it should be the case and also [the spec](./spec/sped.md) suggests so.

This change fixes this while introducing one more database hit than the previous solution.
This is an important hotfix and hence should be released while a more resource friendly solution can be developed.

## How Has This Been Tested?

Added functional test

## Due Diligence

* [ ] Breaking change


xav (approved):

src/handlers/push_message.rs
```diff
@@ -67,7 +79,7 @@ pub async fn handler(
             "notification ({}) already received for client ({})",
             body.id, id
         );
-        return Ok(Response::new_success(StatusCode::ACCEPTED));
+        return Ok(Response::new_success(StatusCode::OK));
```
HarryET, line 82: This should be accepted - we offer no guarantees that it will be delivered

src/handlers/push_message.rs
```diff
@@ -51,6 +51,18 @@ pub async fn handler(
...
+    if let Ok(_notification) = state
+        .notification_store
+        .get_notification(&body.id, &tenant_id)
+        .await
+    {
+        info!(
+            "notification ({}) already received for client ({})",
+            body.id, id
+        );
+        return Ok(Response::new_success(StatusCode::OK));
```
HarryET, line 63: This should be accepted - we offer no guarantees that it will be delivered

HarryET (changes requested):

src/handlers/push_message.rs
arein: It's to distinguish whether we actually did anything or whether it was a dupe

HarryET: Okay!

src/handlers/push_message.rs
```diff
@@ -51,6 +51,18 @@ pub async fn handler(
     let client = state.client_store.get_client(&tenant_id, &id).await?;
     info!("fetched client ({}) for tenant ({})", &id, &tenant_id);
 
+    if let Ok(_notification) = state
```
HarryET, line 54: The notification may be stored but might now have processed we should have a better way to do this eventually!

HarryET (changes requested):
The notification store doesn't store the payloads in the database right now which stops this logic working please update the create or update method to upsert the previous payloads and update the last payload columns

arein:
> 

did you read the PR details where I explicitly explain that this is a hotfix?
I am on-call for this service and it's my weekend. The fix I present meets the requirements and does not stand in the way of your preferred refactor.
You have all means to prioritize your preferred refactor solution through the normal routes that can be achieved during regular work hours now that I am closing the leak.

HarryET (approved):
I was only pointing out that half of this "fix" isn't going to do anything.I've  approved this and have made an issue to fix the problem properly, see: https://github.com/WalletConnect/echo-server/issues/98

PR #99, opened 2023-02-06
fix(hotfix): cors not properly configured

arein:
# Description

Our CORS config doesn't work. This is a hotfix that makes it work.

## How Has This Been Tested?

Added functional tests to confirm the before state was broken and the after state works.

`cargo test`

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@ use {
     tokio::{select, sync::broadcast},
     tower::ServiceBuilder,
     tower_http::{
-        cors::CorsLayer,
+        cors::{AllowOrigin, CorsLayer},
         trace::{DefaultMakeSpan, DefaultOnRequest, DefaultOnResponse, TraceLayer},
     },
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -150,12 +150,13 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
             global_middleware.clone().layer(
                 CorsLayer::new()
                     .allow_methods([Method::GET, Method::POST, Method::DELETE])
-                    .allow_origin(
```
HarryET, line 153: Can you just update the terraform to * instead of a code change for that?

HarryET (approved):

xav (approved):
Can you fix the clippy warning?

src/lib.rs
arein: That didn't work for me when I tried to test it in the functional tests.

arein: I think we'll need to do it slightly different.

PR #103, opened 2023-02-15
feat: metrics

HarryET:
# Description

Adds more metrics as well as a macro for easily incrementing a counter on the `state.metrics` property when they are enabled.

## How Has This Been Tested?

N/a, will test in grafana after deployed to prod and then PR the new dashboard back

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,19 +1,23 @@
 #[macro_export]
 macro_rules! increment_counter {
-    ($state:ident, $metric:ident) => {
-        if let Some(metrics) = &$state {
+    ($state:ident.$property:ident, $metric:ident) => {{
+        use {opentelemetry::Context, tracing::debug};
+
+        if let Some(metrics) = &$state.$property {
             metrics.$metric.add(&Context::current(), 1, &[]);
             debug!("incremented `{}` counter", stringify!($metric));
         }
-    };
+    }};
 }
 
 #[macro_export]
 macro_rules! decrement_counter {
-    ($state:ident, $metric:ident) => {
-        if let Some(metrics) = &$state {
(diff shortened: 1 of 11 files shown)


src/macros.rs
```diff
@@ -0,0 +1,23 @@
...
+macro_rules! increment_counter {
+    ($state:ident.$property:ident, $metric:ident) => {{
+        use {opentelemetry::Context, tracing::debug};
+
+        if let Some(metrics) = &$state.$property {
+            metrics.$metric.add(&Context::current(), 1, &[]);
+            debug!("incremented `{}` counter", stringify!($metric));
+        }
+    }};
+}
```
chris13524, line 11: thought: I was going to suggest using a function instead of a macro here, but looks like the problem with that is not being able to read the metric name from the metric object itself (for the `debug!()`). Interesting limitation of the library.

src/macros.rs
```diff
@@ -0,0 +1,23 @@
+#[macro_export]
+macro_rules! increment_counter {
+    ($state:ident.$property:ident, $metric:ident) => {{
+        use {opentelemetry::Context, tracing::debug};
+
+        if let Some(metrics) = &$state.$property {
```
chris13524, line 6: Looks like this macro only supports a first argument in the form `ident.ident`? What if I wanted to pass a more deeply-nested identifier such as `ident.ident.ident`? Would using `$metrics:expr` as the first macro parameter make this more generalized?

chris13524 (approved):

src/macros.rs
HarryET: It can be added but right now I know that it will only be one deep -> can be looked at later though

PR #106, opened 2023-02-18
fix: apns support

HarryET:
# Description
APNS wasn't setting type so it was always disabled for new staging projects, changes never reached production so no migration needed.

This also splits the updating methods to allow FCM & APNS to be updated independently of each other

## How Has This Been Tested?
N/a

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -40,4 +40,4 @@ jobs:
       - run: yarn integration:$ENVIRONMENT
         env:
           ENVIRONMENT: ${{ inputs.environment }}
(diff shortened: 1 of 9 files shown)


HarryET:
\cc @Cali93 this gives you the `apns_type` for each tenant too

arein (approved):
would be nice to see in the PR name or details what exactly is broken

do we need to migrate anything?

HarryET:
> would be nice to see in the PR name or details what exactly is broken
> 
> do we need to migrate anything?

Shouldn't need any migrations

PR #110, opened 2023-02-28
feat: `client_id` auth

HarryET:
# Description
Adds non-breaking client id/jwt validation on the client endpoints

## How Has This Been Tested?
N/a, non-breaking so will do integration testing with SDK teams once they have upgraded

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/delete_client.rs
+++ b/src/handlers/delete_client.rs
@@ -18,7 +18,7 @@ pub async fn handler(
     StateExtractor(state): StateExtractor<Arc<AppState>>,
     headers: HeaderMap,
 ) -> Result<Response> {
-    if !authenticate_client(headers, |client_id| {
+    if !authenticate_client(headers, &state.config.public_url, |client_id| {
         if let Some(client_id) = client_id {
             &client_id == &body.client_id
         } else {
--- a/src/handlers/delete_client.rs
+++ b/src/handlers/delete_client.rs
@@ -1,5 +1,4 @@
 use {
-    super::register_client::RegisterBody,
     crate::{
         decrement_counter,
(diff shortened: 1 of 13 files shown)


terraform/variables.tf
```diff
@@ -10,7 +10,7 @@ variable "azs" {
 
 variable "public_url" {
   type    = string
-  default = "echo.walletconnect.com"
+  default = "https://echo.walletconnect.com"
```
HarryET: @Rakowskiii I think its the `https://` prefix here causing some of the terraform issues

src/handlers/delete_client.rs
```diff
@@ -1,23 +1,39 @@
...
     Path((tenant_id, id)): Path<(String, String)>,
     StateExtractor(state): StateExtractor<Arc<AppState>>,
+    headers: HeaderMap,
 ) -> Result<Response> {
     let id = id
         .trim_start_matches(DECENTRALIZED_IDENTIFIER_PREFIX)
         .to_string();
 
+    let client_to_be_deleted = ClientId::new(id.clone().into());
+    if !authenticate_client(headers, &state.config.public_url, |client_id| {
```
HarryET, line 28: `public_url` needs the `https://` or `http://` prefix now. Same for all other function calls

Rakowskiii: Wouldn't this line do that? https://github.com/WalletConnect/echo-server/blob/8f592e4253bee4cb34cf4de7563e80da2338f32f/terraform/ecs/main.tf#L73

HarryET: Yeah thats good!

PR #113, opened 2023-03-16
feat: added tracing of JWT claims verification failure

Rakowskiii:
# Description

Log errors in verification of JWT tokens to ensure we are able to quickly debug issues.

## Due Diligence

* [ ] Breaking change


src/handlers/mod.rs
```diff
@@ -29,7 +30,12 @@ where
 {
     return if let Some(auth_header) = headers.get(axum::http::header::AUTHORIZATION) {
         let header_str = auth_header.to_str()?;
-        let client_id = Jwt(header_str.to_string()).decode(&HashSet::from([aud.to_string()]))?;
+        let client_id = Jwt(header_str.to_string())
+            .decode(&HashSet::from([aud.to_string()]))
+            .map_err(|e| {
+                info!("Invalid claims: {:?}", e);
```
HarryET, line 36: Can we get the client id here too so we can search for these logs?

HarryET (changes requested):

src/handlers/mod.rs
Rakowskiii: In here no, as this error may also be due to JWT being invalid -> not carrying client_id. We can log client_id in check closure, but then we'll need to remember to add this everywhere we provide it.

src/handlers/delete_client.rs
```diff
@@ -27,6 +27,10 @@ pub async fn handler(
     dbg!(&state.config.public_url);
     if !authenticate_client(headers, &state.config.public_url, |client_id| {
         if let Some(client_id) = client_id {
+            info!(
```
Rakowskiii, line 30: @HarryET this is what we can do regarding client_id. Would this be sufficient?

src/handlers/mod.rs
HarryET: Ah ok, I guess log lines should be close together so we can figure it out from that

src/handlers/delete_client.rs
HarryET: Yeah, thats good. I think there is a way with log? maybe a `?` syntax not sure but that can always be done later

HarryET (approved):

PR #114, opened 2023-03-17
feat: Analytics

HarryET:
# Description
Adds analytics to be used by the data team

## How Has This Been Tested?
N/a

## Due Diligence

* [ ] Breaking change


src/analytics/message_info.rs
```diff
@@ -0,0 +1,14 @@
+use {parquet_derive::ParquetRecordWriter, serde::Serialize, std::sync::Arc};
+
+#[derive(Debug, Clone, Serialize, ParquetRecordWriter)]
+#[serde(rename_all = "camelCase")]
+pub struct MessageInfo {
```
HarryET, line 5: @biranyucel, here is all the stuff we are exporting for analytics. Look good? Also anything else you could think you want?

biranyucel: Don't we have the time information?

HarryET: We can get the time info! Do you want time recieved?

biranyucel: Yes, please! Also, is there a user_agent here? Can we get that information as well?

HarryET: We could but I'm not sure it will help as the requests recieved here are from the relay 😞

src/handlers/push_message.rs
```diff
@@ -120,5 +124,30 @@ pub async fn handler(
         Provider::Noop(_) => {}
     }
 
+    // Analytics
```
Rakowskiii, line 130: Maybe we could log this dispatch too? In case there are some problems with analytics, we'll be sure that at least we are sending them. What do you think?

Rakowskiii (approved):
Looks good

src/analytics/message_info.rs
biranyucel: Okay, with the client-wallet pair, I think most of that information could be gathered from relay_messages on our side if we need it.    
Based on our chat, your suggestion to have a client registration date was so good. Can we add that, please?

src/handlers/push_message.rs
```diff
@@ -120,5 +124,30 @@ pub async fn handler(
...
+    if let Some(analytics) = &state.analytics {
+        let (country, continent, _) =
+            analytics
+                .geoip
+                .lookup_geo_data(addr.ip())
+                .map_or((None, None, None), |geo| {
+                    // TODO (Harry): Figure out region mapping
```
xav: https://github.com/WalletConnect/rs-relay/blob/d7d57340b69b84221e1418132bb025df7c18325d/src/handler/publish.rs#L450

src/handlers/push_message.rs
```diff
@@ -120,5 +127,34 @@ pub async fn handler(
...
 
+    // Analytics
+    tokio::spawn(async move {
+        if let Some(analytics) = &state.analytics {
+            let (country, continent, region) =
+                analytics
+                    .geoip
+                    .lookup_geo_data(addr.ip())
+                    .map_or((None, None, None), |geo| {
+                        // TODO (Harry): Figure out region mapping
```
xav: You can remove the TODO now :)

src/handlers/register_client.rs
```diff
@@ -79,5 +81,31 @@ pub async fn handler(
...
 
+    // Analytics
+    tokio::spawn(async move {
+        if let Some(analytics) = &state.analytics {
+            let (country, continent, region) =
+                analytics
+                    .geoip
+                    .lookup_geo_data(addr.ip())
+                    .map_or((None, None, None), |geo| {
+                        // TODO (Harry): Figure out region mapping
```
xav: Remove the TODO here too

xav (approved):

PR #119, opened 2023-04-23
feat: logging

HarryET:
# Description
Have better logging in Echo Server to help us Debug 4XX and 5XX requests that we have seen, it should allow us to find issues faster.

Follow-up PR coming to add UA to Analytics to give insights to current versions of SDKs

## How Has This Been Tested?
Tests still passing, check local console and will check console on staging

## Due Diligence

* [ ] Breaking change


arein (approved):
Thank you!

PR #122, opened 2023-04-24
feat: `project_id` validation

### LINKED ISSUE #120: feat: Validate `tenant_id` is a valid `project_id` using Internal API
Check `tenant_id` is a valid `project_id` and reject when it is not valid. This can be done using the [cerberus crate](https://github.com/WalletConnect/cerberus). This is a pre-requisite for the Supabase JWT auth for the Tenant management endpoints

HarryET:
# Description
Resolves #120 

## How Has This Been Tested?
Locally tested.

## Due Diligence

* [x] Breaking change
    Now requires valid ids this should already be the case but is technically breaking

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,10 @@ LOG_LEVEL_OTEL=info,echo-server=trace
 TENANT_DATABASE_URL=
 DEFAULT_TENANT_ID= # This has a default value and dosen't hold much impact to the running of echo-server
 
+# Cloud App
+CLOUD_API_URL=
+CLOUD_API_KEY=
+
 # CORS
 CORS_ALLOWED_ORIGINS=*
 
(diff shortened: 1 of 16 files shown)


HarryET:
How should the API Key be managed?
\cc @xav @arein (security)

arein (approved):
Nice! So Rainbow just doesn't use the `Cloud` feature?

HarryET:
> Nice! So Rainbow just doesn't use the `Cloud` feature?

Yup, they can also remove the multi-tenant stuff all together! A few things are feature flagged now to help people with self-hosting.

arein:
> How should the API Key be managed? \cc @xav @arein (security)

API keys are hardcoded in the Cloud App currently. Ideally you hardcode a new one just for echo else use the existing one (1Password)

HarryET:
> > How should the API Key be managed? \cc @xav @arein (security)
> 
> API keys are hardcoded in the Cloud App currently. Ideally you hardcode a new one just for echo else use the existing one (1Password)

Okay, should I add it as a GH Actions Secret then pass it as a TF Secret Variable?

PR #139, opened 2023-06-19
enhance: Better Error Responses

### LINKED ISSUE #135: Improve error handling
We have `error.rs` which converts our Error enum into HTTP Responses, right now there is a catch all clause which is being called more often than it needs to be.

The implementation of `IntoResponse` should be updated to handle errors in a more descriptive way for end users and the relay.

HarryET:
# Description
Respond with more descriptive errors

Resolves #135 

## How Has This Been Tested?
Ran locally to ensure still builds

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -7,7 +7,6 @@ use {
     },
     axum::response::{IntoResponse, Response},
     hyper::StatusCode,
-    std::fmt::format,
 };
 
 pub type Result<T> = std::result::Result<T, Error>;
@@ -453,8 +452,8 @@ impl IntoResponse for Error {
                 },
             ], vec![]),
             e => {
-                warn!("Error does not have response clause, {:?}", Self);
-                
+                warn!("Error does not have response clause, {:?}", e);
+
                 crate::handlers::Response::new_failure(StatusCode::INTERNAL_SERVER_ERROR, vec![
                     ResponseError {
                         name: "unknown_error".to_string(),
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,3 @@
-use std::fmt::format;
 use {
     crate::{
         handlers::{ErrorField, ErrorLocation, ResponseError},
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -347,17 +347,124 @@ impl IntoResponse for Error {
...
+            ], vec![]),
+            Error::GeoIpReader(_) | Error::BatchCollector(_) => crate::handlers::Response::new_failure(StatusCode::INTERNAL_SERVER_ERROR, vec![
                 ResponseError {
-                    name: "dbg".to_string(),
-                    message: format!("{e:?}"),
-                }
-            ], vec![])
+                    name: "o11y".to_string(),
+                    message: "Internal error monitoring the request".to_string(),
+                },
```
arein, line 440: curious: when is the `GeoIpReader` used?

src/error.rs
```diff
@@ -347,17 +347,124 @@ impl IntoResponse for Error {
...
+            ], vec![]),
+            Error::Registry(_) => crate::handlers::Response::new_failure(StatusCode::INTERNAL_SERVER_ERROR, vec![
+                ResponseError {
+                    name: "internal_api_failed".to_string(),
+                    message: "Please check https://status.walletconnect.com as an internal API failed to resolve this request".to_string(),
+                },
```
arein, line 446: Which is this application not using `cerberus` for this?

arein (approved):

src/error.rs
HarryET: https://github.com/WalletConnect/echo-server/blob/90a3a7e984b2a9e8f3d7ddbbf3c1de42c4919fbe/src/handlers/register_client.rs#L105-L136

src/error.rs
HarryET: `cerberus` is used currently v0.5.0, I'll make a follow-up PR to bump that if applicable

HarryET:
CI/CD failed but it was a known issue at the time, merged the fix for that first and checking ci passed on main, if not will make a follow-up pr

PR #143, opened 2023-06-21
feat: bump parquet and fix builds

HarryET:
# Description
Bumps Parquet and Fixes builds to unblock pipeline

## How Has This Been Tested?
Built image locally

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/analytics/mod.rs
+++ b/src/analytics/mod.rs
@@ -2,19 +2,21 @@ use {
     crate::{
         analytics::{client_info::ClientInfo, message_info::MessageInfo},
         config::Config,
-        error::{Error, Result},
+        error::Result,
         log::prelude::*,
     },
     aws_config::meta::region::RegionProviderChain,
-    aws_sdk_s3::{Client as S3Client, Region},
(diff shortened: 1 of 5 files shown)


src/analytics/mod.rs
```diff
@@ -45,39 +47,37 @@ impl PushAnalytics {
...
-            )
+            let collector = ParquetWriter::<MessageInfo>::new(opts.clone(), exporter)?;
+            Analytics::new(collector)
         };
 
         let clients = {
-            let exporter = AwsExporter::new(AwsExporterOpts {
+            let exporter = AwsExporter::new(AwsOpts {
+                // TODO (Harry Bairstow): Do we need a prefix?
+                export_prefix: "",
```
arein, line 71: did we have one before?

arein (approved):

src/analytics/mod.rs
HarryET: Not for Echo Server, but `gorgon` requires this now and the relay has one, going to investigate with Data team

PR #148, opened 2023-06-24
fix: return correct error code

### LINKED ISSUE #147: fix: constant 5XX errors
5XX and `error` level log seem wrong and should be `4XX` and `warn`. Just a hunch.

HarryET:
# Description
Returns the correct error code when device tokens have expired or are invalid for APNS, requires a follow-up PR to delete clients from the database

Resolves #147 

## How Has This Been Tested?
Responses should improve in relay logs, non-breaking

## Due Diligence

* [ ] Breaking change


arein (approved):
Thank you !

PR #155, opened 2023-07-03
fix(cors-headers): allow authorization and content-type headers

### LINKED ISSUE #87: feat: CORS headers

Cali93:
# Description

Resolves #87

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -187,7 +187,11 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) -> 
                                 .iter()
                                 .map(|v| v.parse::<HeaderValue>().unwrap())
                                 .collect::<Vec<HeaderValue>>(), */
-                        .allow_origin(AllowOrigin::any()),
+                        .allow_origin(AllowOrigin::any())
+                        .allow_headers(vec![
+                            "Content-Type".parse().unwrap(),
+                            "Authorization".parse().unwrap(),
+                        ]),
                 ),
             );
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -187,7 +187,11 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
                                 .iter()
                                 .map(|v| v.parse::<HeaderValue>().unwrap())
                                 .collect::<Vec<HeaderValue>>(), */
-                        .allow_origin(AllowOrigin::any()),
+                        .allow_origin(AllowOrigin::any())
+                        .allow_headers(vec![
+                            "Content-Type".parse().unwrap(),
+                            "Authorization".parse().unwrap(),
+                        ]),
```
Rakowskiii:
```suggestion
                        .allow_headers([hyper::http::header::CONTENT_TYPE, hyper::http::header::AUTHORIZATION]),
```

Rakowskiii: I think this would be a cleaner way to do it 😄

Rakowskiii (approved):

HarryET (changes requested):
Why is the lock file updated in this PR? Otherwise looks good! 😁

PR #161, opened 2023-07-12
feat: more analytics

HarryET:
# Description
Includes requested additional analytics

Resolves # (issue)

## How Has This Been Tested?
Build locally

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/push_message.rs
+++ b/src/handlers/push_message.rs
@@ -1,3 +1,4 @@
+use axum::response::IntoResponse;
 #[cfg(feature = "analytics")]
 use {
     crate::analytics::message_info::MessageInfo,
@@ -23,7 +24,6 @@ use {
     axum::{
         extract::{Json, Path, State as StateExtractor},
         http::{HeaderMap, StatusCode},
-        response::IntoResponse,
     },
     serde::{Deserialize, Serialize},
     std::sync::Arc,
@@ -132,17 +132,17 @@ pub async fn handler(
     Ok(response)
 }
 
-pub async fn handler_internal(
+pub async fn handler_internal<MessageInfo>(
     Path((tenant_id, id)): Path<(String, String)>,
     StateExtractor(state): StateExtractor<Arc<AppState>>,
     headers: HeaderMap,
     RequireValidSignature(Json(body)): RequireValidSignature<Json<PushMessageBody>>,
 ) -> Result<(Response, Option<MessageInfo>)> {
     #[cfg(feature = "analytics")]
-    let topic: Option<Arc<str>> = body.payload.topic.as_ref().map(|t| t.clone().into());
+    let topic: Option<Arc<str>> = body.payload.clone().topic.as_ref().map(|t| t.clone().into());
 
     #[cfg(feature = "analytics")]
-    let (flags, encrypted) = (body.payload.flags, body.payload.is_encrypted());
+    let (flags, encrypted) = (body.payload.clone().flags, body.payload.is_encrypted());
 
     #[cfg(feature = "analytics")]
     let mut analytics = MessageInfo {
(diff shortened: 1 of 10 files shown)


src/handlers/push_message.rs
```diff
@@ -54,13 +55,92 @@ pub async fn handler(
...
+            let res = error.into_response();
+            let status_code = res.status().clone().as_u16();
+            (status_code, res)
+        }
+    };
+
+    #[cfg(feature = "analytics")]
+    if let Some(mut message_info) = analytics_option {
+        message_info.status = status;
+        message_info.success = status >= 200 && status < 300;
```
xav: Nice to have, but isn't it enough with just `status` in the table?

src/handlers/push_message.rs
```diff
@@ -54,13 +55,92 @@ pub async fn handler(
...
+        message_info.response_message = Some(response.body().to_string());
```
xav: Is the presence of 'response_message' the only differentiator between a request and a response?
In that case, can we have a field that tells explicitly which is which to avoid potential downstream issues?

src/handlers/push_message.rs
```diff
@@ -54,13 +55,92 @@ pub async fn handler(
...
+        country: None,
+        continent: None,
+        project_id: tenant_id.into(),
+        client_id: id.into(),
+        topic,
+        push_provider: client.push_type.as_str().into(),
+        encrypted,
+        flags,
+        status: 0,
+        success: false,
```
xav: One more reason to remove this field. Someone someday will just do a count on 'success=false' and panic because there are a lot of failures :)

src/handlers/push_message.rs
```diff
@@ -91,7 +171,18 @@ pub async fn handler(
             last_recieved_at = %notification.last_received_at,
             "notification has already been received"
         );
-        return Ok(Response::new_success(StatusCode::OK));
+
+        #[cfg(feature = "analytics")]
+        {
+            analytics.success = true;
```
xav: Is that a third type of item? In that case, maybe a third type in addition to 'request' and 'response; could cover this?

xav (commented):
Did you check with the data team already about these changes? You need a green light from them before changing the format/semantics of the data lake.

src/handlers/push_message.rs
dnul: my intention with the success field was to give echo server the responsibility to determine when a request is successful or not, given it has more context on what's being done. Given it's just

>        message_info.success = status >= 200 && status < 300;

I don't mind dropping the column. 

Consider if in  the future  things becomes more complicated like `if status == 2xx AND some other condition` then it would be a problem for data to determine, with this column you control whether the outcome was successful or not.

arein (approved):

src/handlers/push_message.rs
HarryET: response message was so we could provide context on errors

src/handlers/push_message.rs
HarryET: Success is redundant as pointed out above, its been removed

src/handlers/push_message.rs
HarryET: done

src/handlers/push_message.rs
HarryET: I've removed it for now

src/handlers/push_message.rs
xav: ok, I think we still need to be able to differentiate between requests and responses

HarryET:
Builds should be all sorted now, waiting for CI & Re-review before merge + deploy

arein (approved):

PR #168, opened 2023-07-27
fix: ensure tenant_id matches the tenant a client is registered with

HarryET:
# Description

While debugging 5XX errors, I noticed there was a potential to send a notification to a client from another tenant. This shouldn't have been as issue as the `push_message` has signature validation and the relay behaviour should have been within specification. 

## How Has This Been Tested?
Builds locally

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -164,7 +164,7 @@ pub enum Error {
     GeoIpS3Failed,
 
     #[error("tenant id and client's registered tenant didn't match")]
-    MissatchedTenantId,
+    MissmatchedTenantId,
 }
 
 impl IntoResponse for Error {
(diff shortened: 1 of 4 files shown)


src/error.rs
```diff
@@ -162,6 +162,9 @@ pub enum Error {
 
     #[error("failed to load geoip database from s3")]
     GeoIpS3Failed,
+
+    #[error("tenant id and client's registered tenant didn't match")]
+    MissatchedTenantId,
```
chris13524:
```suggestion
    MissmatchedTenantId,
```

chris13524 (changes requested):

HarryET:
I noticed something while making this, we were not settings the `tenant_id` for stored tenants, so for now we need to store for all existing and not change behaviour for past. When we re-build for Push Service, client's won't be able to be migrated so this can be solved there.

arein (approved):

PR #169, opened 2023-08-01
feat: fcm verification

HarryET:
# Description
FCM API Keys are not validated to ensure they work with the FCM Api before the tenant is saved.

Resolves # (issue)

## How Has This Been Tested?
Adding interation tests to ensure it fails on invalid API Keys, and passes on valid ones. Requires valid FCM API Key in the env vars

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -3,7 +3,7 @@ let
 in
   with nixpkgs;
   stdenv.mkDerivation {
-    name = "music-reader-env";
+    name = "echo-server";
     buildInputs = [
       cargo
       rustc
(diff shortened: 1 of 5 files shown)


arein (approved):
Nice! Do we need to do something with existing invalid keys ?

HarryET:
> Nice! Do we need to do something with existing invalid keys ?

I'm investigating maybe a cron job that checks once a day for invalid api keys and then suspends the projects, this would then also catch projects which invalidate their api key after submitting it

PR #172, opened 2023-08-02
feat: Web3Inbox CORS Support

HarryET:
# Description
Adds CORS headers for W3I's FCM Web Push use case

## How Has This Been Tested?
N/a

## Due Diligence

* [ ] Breaking change


src/lib.rs
```diff
@@ -238,6 +249,17 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
             post(handlers::single_tenant_wrappers::push_handler),
         )
         .layer(global_middleware)
+        .layer(
+            global_middleware.clone().layer(
+                CorsLayer::new()
+                    .allow_methods([Method::POST, Method::DELETE])
```
devceline: I haven't written a CORS policy in a while, but don't we need to add `Method:OPTIONS`?

HarryET: This should be fine, it's the same way it's written in cast and above in echo

devceline (approved):
LGTM

arein (approved):

PR #174, opened 2023-08-02
feat: APNS Verification

HarryET:
# Description
Verify provided APNS credentials are valid

## How Has This Been Tested?
New tests added for invalid tokens/certs, still needs tests for valid but that requires injecting files/tokens for apns

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/multitenant/apns.rs
+++ b/tests/functional/multitenant/apns.rs
@@ -1,0 +1,118 @@
+use {
+    crate::context::EchoServerContext,
+    echo_server::handlers::create_tenant::TenantRegisterBody,
+    random_string::generate,
+    test_context::test_context,
+};
+
+// #[test_context(EchoServerContext)]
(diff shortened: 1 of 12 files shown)


tests/functional/multitenant/apns.rs
```diff
@@ -0,0 +1,118 @@
...
+//         .multipart(&form)
+//         .send()
+//         .await
+//         .expect("Call failed");
+//
+//     assert!(
+//         response.status().is_success(),
+//         "Response was not successful"
+//     );
+// }
```
arein: is this a leftover?

arein (approved):

tests/functional/multitenant/apns.rs
HarryET: It needs to be added back once we have a good way to pass valid APNS credentials into the tests

PR #175, opened 2023-08-07
feat: always include topic data for push

HarryET:
# Description
Request from Family, always include the topic data that Echo Server receives as additional data on APNS/FCM requests

Resolves `n/a`

## Due Diligence

* [ ] Breaking change


arein (approved):
Tests missing?

PR #177, opened 2023-08-08
feat: Suspend Broken Tenants & Delete Broken Clients

HarryET:
# Description
Saves IP reputation and removes random 5XX errors

## How Has This Been Tested?
N/a

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/migrations/1691518766_add-suspension.sql
+++ b/migrations/1691518766_add-suspension.sql
@@ -1,0 +1,5 @@
+alter table public.tenants
+    add suspended bool not null default false;
+
+alter table public.tenants
+    add suspended_reason text;
--- a/src/error.rs
+++ b/src/error.rs
@@ -171,6 +171,12 @@ pub enum Error {
 
     #[error("invalid apns creds")]
     BadApnsCredentials,
+
+    #[error("invalid device token")]
+    ClientDeleted,
+
+    #[error("invalid tenant configuration")]
+    TenantSuspended,
 }
 
 impl IntoResponse for Error {
@@ -525,6 +531,18 @@ impl IntoResponse for Error {
                     location: ErrorLocation::Path,
                 }
             ]),
+            Error::ClientDeleted => crate::handlers::Response::new_failure(StatusCode::ACCEPTED, vec![
+                ResponseError {
+                    name: "client_deleted".to_string(),
+                    message: "Request Accepted, client deleted due to invalid token".to_string(),
+                },
+            ], vec![]),
+            Error::TenantSuspended => crate::handlers::Response::new_failure(StatusCode::ACCEPTED, vec![
+                ResponseError {
+                    name: "tenant_suspended".to_string(),
+                    message: "Request Accepted, tenant suspended due to invalid configuration".to_string(),
+                },
+            ], vec![]),
             e => {
                 warn!("Error does not have response clause, {:?}", e);
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,6 +37,9 @@ pub enum Error {
     #[error(transparent)]
     Fcm(#[from] fcm::FcmError),
 
+    #[error("FCM Responded with an error")]
+    FcmResponse(fcm::ErrorReason),
+
     #[error(transparent)]
     Io(#[from] std::io::Error),
 
@@ -172,10 +175,10 @@ pub enum Error {
     #[error("invalid apns creds")]
     BadApnsCredentials,
 
-    #[error("invalid device token")]
+    #[error("client deleted due to invalid device token")]
     ClientDeleted,
 
-    #[error("invalid tenant configuration")]
+    #[error("tenant suspended due to invalid configuration")]
     TenantSuspended,
 }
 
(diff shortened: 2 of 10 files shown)


src/handlers/push_message.rs
```diff
@@ -358,10 +362,29 @@ pub async fn handler_internal(
...
 
-    provider
-        .send_notification(client.token, body.payload)
-        .await
-        .map_err(|e| (e, analytics.clone()))?;
+    match provider.send_notification(client.token, body.payload).await {
+        Ok(_) => Ok(()),
+        Err(error) => match error {
+            Error::BadDeviceToken => {
+                state.client_store.delete_client(&tenant_id, &id);
```
arein: is this a soft delete?

arein (approved):
LGTM other than tests are missing here as well.
Would be good to cover in functional tests

src/stores/tenant.rs
```diff
@@ -373,6 +378,18 @@ impl TenantStore for PgPool {
 
         Ok(res)
     }
+
+    async fn suspend_tenant(&self, id: &str, reason: &str) -> Result<()> {
+        sqlx::query_as::<sqlx::postgres::Postgres, Tenant>(
+            "UPDATE public.tenants SET suspended = true, suspended_reason = $2::text WHERE id = \
+             $1 RETURNING *;",
```
chris13524: I think `RETURNING *` is not needed?

```suggestion
             $1;",
```

src/providers/fcm.rs
```diff
@@ -55,10 +56,30 @@ impl PushProvider for FcmProvider {
...
+                            _ => Ok(()),
+                        }
+                    } else {
+                        Ok(())
+                    }
+                }
+            },
+            Err(e) => match e {
+                FcmError::Unauthorized => Err(Error::BadFcmApiKey),
+                _ => Ok(()),
```
chris13524: Is `Ok(())` the right behavior here? Shouldn't we propagate `e`?

src/providers/fcm.rs
```diff
@@ -55,10 +56,30 @@ impl PushProvider for FcmProvider {
...
+        match result {
+            Ok(val) => match val {
+                FcmResponse { error, .. } => {
+                    if let Some(error) = error {
+                        match error {
+                            ErrorReason::MissingRegistration
+                            | ErrorReason::InvalidRegistration
+                            | ErrorReason::NotRegistered => Err(Error::BadDeviceToken),
+                            ErrorReason::InvalidApnsCredential => Err(Error::BadApnsCredentials),
+                            _ => Ok(()),
```
chris13524: Should we propagate `error` instead of returning `Ok(())`?

src/stores/tenant.rs
```diff
@@ -435,4 +452,8 @@ impl TenantStore for DefaultTenantStore {
     ) -> Result<Tenant> {
         panic!("Shouldn't have run in single tenant mode")
     }
+
+    async fn suspend_tenant(&self, id: &str, reason: &str) -> Result<()> {
+        panic!("Shouldn't have run in single tenant mode")
```
chris13524, line 478: Error message could be improved:

```suggestion
        panic!("Invalid APNS or FCM credentials. Cannot suspend tenant in single-tenant mode, so panicking instead.")
```

src/error.rs
```diff
@@ -171,6 +171,12 @@ pub enum Error {
 
     #[error("invalid apns creds")]
     BadApnsCredentials,
+
+    #[error("invalid device token")]
+    ClientDeleted,
+
+    #[error("invalid tenant configuration")]
```
chris13524:
```suggestion
    #[error("tenant suspended due to invalid configuration")]
```

src/error.rs
```diff
@@ -171,6 +171,12 @@ pub enum Error {
 
     #[error("invalid apns creds")]
     BadApnsCredentials,
+
+    #[error("invalid device token")]
```
chris13524:
```suggestion
    #[error("client deleted due to invalid device token")]
```

migrations/1691518766_add-suspension.sql
```diff
@@ -0,0 +1,5 @@
+alter table public.tenants
+    add suspended bool not null default false;
+
+alter table public.tenants
+    add suspended_reason text;
```
chris13524, line 5: Looks like we don't read `suspended_reason` anywhere after writing it, so assuming this is for debugging/audit purposes only for now.

chris13524 (changes requested):

src/stores/tenant.rs
HarryET: There error message is the same for all the functions, they cannot be run in single-tenant mode

migrations/1691518766_add-suspension.sql
HarryET: We read it when we query the tenant and send it back to the cloud app to be displayed

src/handlers/push_message.rs
HarryET: No its a hard delete as the data we have stored is now useless, we still have data about the client in datalake

src/stores/tenant.rs
HarryET: good catch, removed

src/stores/tenant.rs
chris13524: Maybe the other functions' error messages could be improved too?

I just don't see any debugging info that would help a service maintainer know that their credential had expired.

chris13524 (approved):

src/stores/tenant.rs
HarryET: wdym? The panics here are for function calls that won't/can't be done in single-tenant mode

chris13524: Ah, if they are impossible to call, then you should use [`unreachable!()`](https://doc.rust-lang.org/std/macro.unreachable.html) instead of `panic!()`

PR #182, opened 2023-08-24
feat: handle invalid topic for token

HarryET:
# Description
Removes 5XX for invalid token with APNS, this could also be a tenant configuration issue however it is returned as a bad device issue due to the fact we can't check if the tenant has a wrongly configured topic on APNS

## How Has This Been Tested?
N/a, should notice 5XX dissapear after deploy

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1100,7 +1100,7 @@ checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "echo-server"
-version = "0.32.0"
+version = "0.33.0"
 dependencies = [
  "a2",
  "async-recursion",
--- a/src/providers/apns.rs
+++ b/src/providers/apns.rs
@@ -103,6 +103,8 @@ impl PushProvider for ApnsProvider {
                     None => Err(Error::Apns(a2::Error::ResponseError(res))),
                     Some(response) => match response.reason {
                         ErrorReason::BadDeviceToken => Err(Error::BadDeviceToken),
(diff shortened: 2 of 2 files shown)


chris13524 (commented):
> could also be a tenant configuration issue

Could? Sounds like this is the _reason_. How would this happen, client bug, perhaps only from old clients?

I think this at least deserves a different error than BadDeviceToken for visibility. Perhaps simply DeviceTokenNotForTopic?

HarryET:
> > could also be a tenant configuration issue
> 
> Could? Sounds like this is the _reason_. How would this happen, client bug, perhaps only from old clients?

It can be caused by two cases:
1. The tenant config is correct and the token given is for the wrong topic
2. The tenant is configured with the wrong topic and the token is the correct topic

The only way to tell with certainty is to test a set of the tokens and figure out if they all error or just this one but that isn't really feasible here so I think the best option is just presume the client is wrong.

> I think this at least deserves a different error than BadDeviceToken for visibility. Perhaps simply DeviceTokenNotForTopic?

I think BadDeviceToken is fine because of the presumption, we are in fact saying it was a bad device token we were given

chris13524 (approved):
I think I understand now. Either the tenant was mis-configured with the wrong topic (from APNS?) or the wallet client has a bug and registering the token on the wrong topic. Better to assume wallet client has the bug since it's more third-party than the tenant is.

Might be good to log this case if it isn't already being logged (so we can notice a mis-configured tenant) but LGTM!

arein (approved):

PR #195, opened 2023-09-14
fix: compilation errors with `functional_tests` flag

geekbrother:
# Description

When the `--features functional_tests` flag is enabled there are few errors that break the building process:

```
error[E0308]: mismatched types
  --> tests/functional/singletenant/push.rs:62:16
   |
62 |         topic: Some(topic.into()),
   |                ^^^^^^^^^^^^^^^^^^ expected `String`, found `Option<_>`
   |
   = note: expected struct `std::string::String`
                found enum `Option<_>`

error[E0308]: mismatched types
  --> tests/functional/stores/notification.rs:38:20
   |
38 |             topic: None,
   |                    ^^^^ expected `String`, found `Option<_>`
   |
   = note: expected struct `std::string::String`
                found enum `Option<_>`
```

Seems that functional tests are out of sync with recent changes. To fix that I've changed fields to the expected types in the errors to make them in sync with the recent changes and make the functional tests compile.

Resolves #194 

## How Has This Been Tested?

It was tested by running `cargo test --features functional_tests`. 
The expected result is successful test building with the `functional_tests` enabled without errors.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #196, opened 2023-09-14
fix: duplicated push tokens

### LINKED ISSUE #193: fix: device token duplication in the `clients` table
**Problem description:**

When an app is reinstalled, the device ID (table `clients` -> column `id`) is changed, but the token remains the same causing the same token to be inserted for a second device ID as `clients` -> `device_token`.
The problem is caused by the SQL query in [client.rs](https://github.com/WalletConnect/echo-server/blob/95474d7a25a8667d129b6660d6a575753ecd938d/src/stores/client.rs#L27-L49):
```sql
INSERT INTO public.clients (id, ... , device_token) ... ON CONFLICT (id) DO UPDATE SET device_token = EXCLUDED.device_token ...
```
In this query on conflicting `id`, we update the `device_token`. To fix the duplication problem we should also do the same on the `device_token`: on conflicting `device_token` update `id`.

**Testing behavior:**

1. Insert a new client `id` and `device_token` where neither exists, expecting a new row to be added.
2. Insert a new token with the existing client `id`, expecting `device_token` to be replaced.
3. Insert new client `id` with same `device_token`, expecting client `id` to be replaced.

```[tasklist]
### Tasks
- [x] [client.rs](https://github.com/WalletConnect/echo-server/tree/95474d7a25a8667d129b6660d6a575753ecd938d/src/stores/client.rs): Change the SQL `INSERT` query in the `create_client` implementation,
- [x] Testing: Add/modify tests for the updated behavior.
```

geekbrother:
# Description

When an app is reinstalled, the device ID (table clients -> column id) is changed, but the token remains the same causing the same token to be inserted for a second device ID as clients -> device_token. 

To make upset for both `id` and `device_token` following changes are made:
- The upset query for the client creation was changed from using `ON CONFLICT` to `DELETE`  and `INSERT` wrapped into a transaction because unfortunately, we can't use `ON CONFLICT` with the exclusion conditions.
- The `client_upsert` functional test was changed to test the new behavior in `create_client` in the following way:
  1. Insert a new client `id` and `device_token` where neither exists, expecting a new row to be added.
  2. Insert a new token with the existing client `id`, expecting `device_token` to be replaced.
  3. Insert new client `id` with same `device_token`, expecting client `id` to be replaced.

Resolves #193

The following migration changes were extracted from this issue into #201 :
> - To change the current behavior to upset for both `id` and `device_token` the second `UNIQUE` constraint for `device_token` is added to the SQL migrations.
> - Removing all duplicating `device_token` records and leaving the one recent is implemented as a SQL script in migrations before the `UNIQUE` constraint for `device_token` is added.

## How Has This Been Tested?

**To fix the building errors for the `--features functional_tests` flag please apply or cherry-pick the #195 first.**

Changes are covered by the `client_upsert_id` and `client_upsert_token` functional tests included in this PR.

To test it locally start the PostgreSQL local instances:
```
docker run \
  -p 5432:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust \
  --name echo-server-pg \
  -d postgres

docker run \
  -p 5433:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust \
  --name echo-server-tenant-pg \
  -d postgres
```

Build and run the `client_upsert_id` and `client_upsert_token` tests:

```
cargo t --features functional_tests functional::stores::client::client_upsert_id  -- --exact
cargo t --features functional_tests functional::stores::client::client_upsert_token  -- --exact
```

The expected result is tests should pass.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/stores/client.rs
+++ b/tests/functional/stores/client.rs
@@ -56,30 +56,77 @@ async fn client_creation_apns(ctx: &mut StoreContext) {
 
 #[test_context(StoreContext)]
 #[tokio::test]
-async fn client_upsert(ctx: &mut StoreContext) {
+async fn client_upsert_token(ctx: &mut StoreContext) {
     let id = gen_id();
 
-    let res = ctx
-        .clients
-        .create_client(TENANT_ID, &id, Client {
-            tenant_id: TENANT_ID.to_string(),
-            push_type: ProviderKind::Apns,
-            token: TOKEN.to_string(),
-        })
-        .await;
-
-    assert!(res.is_ok());
-
-    let upsert_res = ctx
-        .clients
+    // Initial Client creation
+    ctx.clients
         .create_client(TENANT_ID, &id, Client {
             tenant_id: TENANT_ID.to_string(),
             push_type: ProviderKind::Fcm,
             token: TOKEN.to_string(),
         })
-        .await;
+        .await
+        .unwrap();
+    let insert_result = ctx.clients.get_client(TENANT_ID, &id).await.unwrap();
+    assert_eq!(insert_result.token, TOKEN);
 
-    assert!(upsert_res.is_ok());
+    // Updating token for the same id
+    let updated_token = gen_id();
+    ctx.clients
+        .create_client(TENANT_ID, &id, Client {
+            tenant_id: TENANT_ID.to_string(),
+            push_type: ProviderKind::Apns,
+            token: updated_token.clone(),
+        })
+        .await
+        .unwrap();
+    let updated_token_result = ctx.clients.get_client(TENANT_ID, &id).await.unwrap();
+    assert_eq!(updated_token_result.token, updated_token);
+
+    // Cleaning up records
+    ctx.clients.delete_client(TENANT_ID, &id).await.unwrap();
+}
+
+#[test_context(StoreContext)]
+#[tokio::test]
+async fn client_upsert_id(ctx: &mut StoreContext) {
+    let id = gen_id();
+
+    // Initial Client creation
+    ctx.clients
+        .create_client(TENANT_ID, &id, Client {
+            tenant_id: TENANT_ID.to_string(),
+            push_type: ProviderKind::Fcm,
+            token: TOKEN.to_string(),
+        })
+        .await
+        .unwrap();
+    let insert_result = ctx.clients.get_client(TENANT_ID, &id).await.unwrap();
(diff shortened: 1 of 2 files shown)


geekbrother:
cc @chris13524 we spoke that it's good to test migration scripts as well. 
When invoking the functional tests the [migrations are applied](https://github.com/WalletConnect/echo-server/blob/95474d7a25a8667d129b6660d6a575753ecd938d/tests/context/stores.rs#L26) so these changes seem tested okay by invoking the `functional::stores::client::client_upsert` test.

tests/functional/stores/client.rs
```diff
@@ -59,27 +59,69 @@ async fn client_creation_apns(ctx: &mut StoreContext) {
...
+                token: TOKEN.to_string(),
+            },
+        )
+        .await
+        .unwrap();
+    let insert_result = ctx.clients.get_client(TENANT_ID, &id).await.unwrap();
+    assert_eq!(insert_result.token, TOKEN);
+    assert_eq!(insert_result.push_type, ProviderKind::Fcm);
+
+    // Updating token for the same id
```
chris13524, line 74: Should each of these scenarios be separate test cases?

xav (approved):
Awesome!

migrations/1694694520_add_token_unique_constrain.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE public.clients
```
chris13524: Shouldn't this be within the same transaction as the delete above?

Also, I think we need to only do these migrations in a second PR after the code changes to avoid getting UNIQUE constraint violations?

chris13524 (changes requested):

tests/functional/stores/client.rs
geekbrother: That makes sense, I'll separate them asap and push changes.

geekbrother: The test is split into `client_upsert_id` and `client_upsert_token`.

migrations/1694694520_add_token_unique_constrain.sql
geekbrother: I think it makes sense, I've extracted the migration to #201 (issue) / #202 (PR) by combining it into the one and wrapping it into a transaction.

chris13524 (approved):

PR #202, opened 2023-09-18
fix: adding SQL migration for making `device_token` UNIQUE

### LINKED ISSUE #201: fix: adding SQL migration file to make `device_token` unique
This issue is a follow-up for #193 where we were extracting the migration into a separate issue and PR.

```[tasklist]
### The new SQL migration file should do the following
- [x] Migration: Clear the `clients` table with duplicate `device tokens` keeping the latest one,
- [x] Migration: Adding `UNIQUE` constraint to the `device_token` column in the `clients` table,
```

geekbrother:
# Description

This PR is a follow-up for #196 with the following SQL migration changes wrapped into the transaction by adding the new `1695024779_add-unique-device-tokens` migration file:
1. Clear the clients table with duplicate device tokens keeping the latest one,
2. Adding UNIQUE constraint to the device_token column in the client's table.

Resolves #201 

## How Has This Been Tested?

Apply or cherry-pick changes from #211 and #213.
When running the functional tests [migrations are applied](https://github.com/WalletConnect/echo-server/blob/95474d7a25a8667d129b6660d6a575753ecd938d/tests/context/stores.rs#L26) that's why we can rely on our upsert functional tests from #196 to test that:
- SQL migration in the new file is correct and runs successfully,
- Changes from the SQL migration don't break the current logic.

To run functional tests start the PostgreSQL instances:

```
docker run \
  -p 5432:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust \
  --name echo-server-pg \
  -d postgres

docker run \
  -p 5433:5432 \
  -e POSTGRES_HOST_AUTH_METHOD=trust \
  --name echo-server-tenant-pg \
  -d postgres
```

Call tests:

```
cargo t --features functional_tests
```

The expected behavior tests should pass.

## Merging policy and dependency

This PR depends on and should be merged after #213 

## Due Diligence

* [ ] Breaking change


xav (approved):

chris13524 (approved):

migrations/1695024779_add-unique-device-tokens.sql
```diff
@@ -0,0 +1,11 @@
+BEGIN;
```
chris13524: Actually I don't think we need to make a transaction ourselves. I think sqlx will do this for us automatically: https://github.com/launchbadge/sqlx/blob/35ba8e2b93f085bbd3b40d31c2f7b01a9911a9de/sqlx-mysql/src/migrate.rs#L175

geekbrother: Thanks for catching it! I'll remove the transaction from the file.

chris13524 (approved):

geekbrother:
Updated the migration filename (increased timestamp) to make it applied after #211

PR #159, opened 2023-07-06
158 move analytics to proper s3 bucket

Rakowskiii:
# Description

Moving echo analytics to shared s3 bucket 

Resolves #158 

## How Has This Been Tested?

- not tested yet

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #204, opened 2023-09-18
fix: removing of the deprecated S3 analytics bucket

### LINKED ISSUE #203: fix: removing of the deprecated analytics S3 bucket from terraform
There is a deprecated analytics S3 bucket in the Terraform that should be removed after the data is migrated.
At the moment the deprecated bucket [making an error](https://github.com/WalletConnect/echo-server/actions/runs/6222304694/job/16887073587#step:8:150) while applying the Terraform config into staging and production. 
The deprecated buckets are removed by [#517](https://github.com/WalletConnect/data-lake/pull/517) from the data lake. The [deprecated bucket](https://github.com/WalletConnect/echo-server/blob/b6c88f19d779bc56a15e73ea85290cff664fc8b9/terraform/analytics/main.tf#L1-L4) should be removed from the `echo-server` Terraform analytics as well.

geekbrother:
# Description

Removing deprecated AWS S3 bucket and config from the Terraform analytics. 
This fixing [the applying error](https://github.com/WalletConnect/echo-server/actions/runs/6222304694/job/16887073587#step:8:150) because we were moved to a single data lake bucket (prev #159).
The bucket was removed from the data lake itself in the PR [#517](https://github.com/WalletConnect/data-lake/pull/517).

Resolves #203 

Tested locally only by validation check `terraform validate`.

## Due Diligence

* [ ] Breaking change


dnul (commented):
will this work?  looks like on line 7 we're referrencing the resource being removed.  I think we should remove everything related to the analytics-data-lake_bucket

terraform/analytics/main.tf
```diff
@@ -1,8 +1,3 @@
-### TODO: This should be removed once data is migrated
-resource "aws_s3_bucket" "analytics-data-lake_bucket" {
-  bucket = "walletconnect.${var.app_name}.${var.environment}.analytics.data-lake"
-}
-
 resource "aws_s3_bucket_acl" "analytics-data-lake_acl" {
   bucket = aws_s3_bucket.analytics-data-lake_bucket.id
```
chris13524: `aws_s3_bucket` here refers to the resource you deleted. I believe this will fail to apply

geekbrother:
> will this work? looks like on line 7 we're referrencing the resource being removed. I think we should remove everything related to the analytics-data-lake_bucket

> aws_s3_bucket here refers to the resource you deleted. I believe this will fail to apply

I've missed the elephant in the room 🤦🏻
Seems that the entire `terraform/analytics` folder and its reference in the `main.tf` should be removed as [we have a unified one](https://github.com/WalletConnect/echo-server/blob/b6c88f19d779bc56a15e73ea85290cff664fc8b9/terraform/main.tf#L15) S3 bucket for analytics in the data lake right now, and the rest seems not relevant anymore. It seems that first changes to remove it started at #159.

I've updated this PR with the rest.

Also, `terraform validate` seems happy right now.

terraform/analytics/main.tf
geekbrother: I've missed the elephant in the room 🤦🏻

xav (approved):

chris13524 (approved):

PR #206, opened 2023-09-21
chore: fix byzantine failures on functional tests

### LINKED ISSUE #205: chore: byzantine faults in CI functional unit tests
When functional unit tests are on there are byzantine faults that occur sometimes: sometimes functional tests fail and pass successfully after a re-run of the CI workflow.
I assume that's because there are no cleanups after each test. 
Also, to track the error messages we should remove `assert!(***.is_ok())` and change to `unwrap()`, because when unwrap panic occurs it prints the complete error information.

geekbrother:
# Description

To fix byzantine failures on functional tests following changes are made:
- Make `TOKEN` [unique for each test](https://github.com/WalletConnect/echo-server/pull/206/commits/7a8fc2c50084803ee3e485dc127457377b98188a) for running tests in parallel and prevent data race conditions in the test PG database,
- Fixing warnings in functional unit tests [about unused variables](https://github.com/WalletConnect/echo-server/pull/206/commits/047facd9091faf4a8e75d4537c32da8d06d473c1).

Resolves #205

## How Has This Been Tested?

Automatically tests by the CI unit test workflow by iterating it 3-5 times.
The expected result is passing the CI unit tests all of the time.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/stores/client.rs
+++ b/tests/functional/stores/client.rs
@@ -7,78 +7,103 @@ use {
     test_context::test_context,
 };
 
-pub const TOKEN: &str = "noop-111-222-333";
-
 #[test_context(StoreContext)]
 #[tokio::test]
 async fn client_creation(ctx: &mut StoreContext) {
-    let res = ctx
-        .clients
-        .create_client(TENANT_ID, &gen_id(), Client {
-            tenant_id: TENANT_ID.to_string(),
-            push_type: ProviderKind::Noop,
-            token: TOKEN.to_string(),
-        })
-        .await;
-
-    assert!(res.is_ok())
+    let id = format!("id-{}", gen_id());
+    let token = format!("token-{}", gen_id());
+    ctx.clients
+        .create_client(
+            TENANT_ID,
+            &id,
+            Client {
+                tenant_id: TENANT_ID.to_string(),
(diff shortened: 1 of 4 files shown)


tests/functional/stores/client.rs
```diff
@@ -7,72 +7,77 @@ use {
...
-            token: TOKEN.to_string(),
+            token,
         })
-        .await;
-
-    assert!(res.is_ok())
+        .await
+        .unwrap();
+    // Cleaning up records
+    ctx.clients.delete_client(TENANT_ID, &id).await.unwrap();
```
chris13524, line 24: Any concern if the `unwrap()` fails and then the records don't get cleaned up?

chris13524 (approved):

tests/functional/stores/client.rs
geekbrother: We are spinning new PostgreSQL docker instances each time we are running the unit tests, so records from the old run should be wiped anyway. I've changed the `TOKEN` to become unique for every test case to fix cases when one case fails/unwrap and solve the domino effect as every case has its own TOKEN in the database table. The main issue was that there was one constant TOKEN for all cases that led to the domino effect and race effect in case run tests in parallel.

PR #214, opened 2023-09-27
fix: adding `id` for the updated release script

### LINKED ISSUE #199: chore: `Cargo.lock` package version out of sync when release CI job is invoked
When the [release](https://github.com/WalletConnect/echo-server/blob/main/.github/workflows/release.yml) CI job is invoked there is a step: `Update version in Cargo.toml` to bump the `echo-server` package version.
The [bash commands in this step](https://github.com/WalletConnect/echo-server/blob/95474d7a25a8667d129b6660d6a575753ecd938d/.github/workflows/release.yml#L44-L47) change the version only in the `Cargo.toml` file and omit `Cargo.lock` file which results that versions in `Cargo.toml` and `Cargo.lock` being out of sync.

geekbrother:
# Description

This PR fixing a nit: `id` for the updated `release` CI workflow script was missing causing the [following error when the release image is building](https://github.com/WalletConnect/echo-server/actions/runs/6318288511/job/17171976388#step:10:223):
```
RROR: tag is needed when pushing to registry
Error: buildx failed with: ERROR: tag is needed when pushing to registry
```

Resolves #199

## How Has This Been Tested?

Should be tested in the GitHub CI workflow running release action.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,6 +29,7 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Update and commit the release version
+        id: release
         uses: WalletConnect/actions/github/update-rust-version/@2.1.4
         with:
           token: ${{ secrets.RELEASE_PAT }}


geekbrother:
cc @xav I think I can merge it now because it's just a nit and doesn't affect any logic.

PR #218, opened 2023-09-27
chore: extra assert APNS response is non-error

### LINKED ISSUE #217: chore: extra assert APNS response is OK

chris13524:
# Description

We are ignoring the response here as a result of poor interface design on the a2 library. While this shouldn't be an issue (I checked the a2 code) I want to assert the response is non-error anyway in-case we are hiding something.

Second, I changed `Ok(_)` to `Ok(())` to make it obvious we aren't ignoring anything.

Resolves #217

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/push_message.rs
+++ b/src/handlers/push_message.rs
@@ -358,7 +358,7 @@ pub async fn handler_internal(
     );
 
     match provider.send_notification(client.token, body.payload).await {
-        Ok(_) => Ok(()),
+        Ok(()) => Ok(()),
         Err(error) => match error {
             Error::BadDeviceToken => {
                 state
--- a/src/providers/apns.rs
+++ b/src/providers/apns.rs
@@ -8,7 +8,7 @@ use {
     a2::{ErrorReason, NotificationBuilder, NotificationOptions},
     async_trait::async_trait,
     std::io::Read,
-    tracing::span,
+    tracing::{span, warn},
 };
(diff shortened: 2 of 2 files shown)


geekbrother (approved):

src/handlers/push_message.rs
```diff
@@ -358,7 +358,7 @@ pub async fn handler_internal(
     );
 
     match provider.send_notification(client.token, body.payload).await {
-        Ok(_) => Ok(()),
+        Ok(()) => Ok(()),
```
xav, line 361: Wouldn't it be better with just `map_err`?

src/providers/apns.rs
```diff
@@ -99,7 +99,18 @@ impl PushProvider for ApnsProvider {
         };
 
         match result {
-            Ok(_) => Ok(()),
+            Ok(response) => {
+                if response.error.is_some() {
+                    warn!(
+                        "Unexpected APNS error. a2 lib shouldn't allow returning Ok containing \
```
xav, line 105: At least we're lucky they're not using `Error 200` :)

xav (approved):

src/handlers/push_message.rs
chris13524: Yeah probably

PR #226, opened 2023-09-29
chore: deploy to production became optional with choice for the `image_tag`

### LINKED ISSUE #224: chore: extract deploying to staging in `cd` workflow
At the moment we can deploy only to the staging and prod environment **together** in our `cd` and `auto_deploy` workflows without the choice to deploy **only to the staging** environment. 
Also, `auto-deploy` is triggered on merge and running deploying to the prod (without running `release`, [example](https://github.com/WalletConnect/echo-server/actions/runs/6350500400)), so every merge the release is [deployed to the prod without running release itself.](https://github.com/WalletConnect/echo-server/blob/00eaa3b3698938754f0f5285674a9a7f150c0a9a/.github/workflows/cd.yml#L10)

My proposal is:
- `auto_deploy` on merging will deploy **only to the staging**,
- `cd` can be triggered manually with the choice to deploy on the staging only or to both, as we [already have in blockchain-api](https://github.com/WalletConnect/blockchain-api/actions/workflows/cd.yaml),
- `release` triggering only manually and running `cd` with deploying to staging and prod.

geekbrother:
# Description

This PR changes the `release` and `cd` GitHub Actions workflows to the following:

Release:

- Adding user inputs for the `release` for options to deploy to `staging` and/or `prod` environments after the release is made.

CD:

- Making deployment to the `staging` and `prod` optional for the `cd` workflow by adding the ability to choose to which environment the image should be deployed when `cd` is invoked.
- Adding `image_tag` string input for the `cd` to make the ability to choose which image version/tag from ECR should be deployed. The default blank input value is the latest release version/tag. This adds an ability to test staging releases by deploying into staging only and falls back to the previous version in case we need to roll it back without making a revert commit and make another release.

Resolves #224

## How Has This Been Tested?

Tested by the [GitHub Actions VSCode plugin](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions) by checking validations.
Smoke testing in a [forked repo](https://github.com/geekbrother/echo-server).
Should be tested when merged as well.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,15 +1,32 @@
 name: cd
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - "main"
-    paths:
-      - "terraform/**"
-  release:
-    types: ["published"]
+    inputs:
+      deploy_to_staging:
+        description: "Deploy to staging"
+        type: boolean
+        required: true
+        default: true
+      deploy_to_prod:
+        description: "Deploy to production"
+        type: boolean
+        required: true
+        default: false
+      image_tag:
+          description: "App image tag. Default: latest release"
+          type: string
+          required: false
+          default: ""
   workflow_call:
     inputs:
+      deploy_to_staging:
+        type: boolean
+        required: true
+        default: true
+      deploy_to_prod:
+        type: boolean
+        required: false
+        default: false
       image_tag:
         type: string
         required: true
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -21,41 +38,12 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
-  get-version:
```
geekbrother, line 24: We don't need this anymore, because we can get the latest release version from the input `image_tag` when calling from the `release`, or it's getting [by the terraform itself when passing an empty one](https://github.com/WalletConnect/echo-server/blob/40e6d51954c798ca86649c1f3c5a1f646523a73c/terraform/main.tf#L8-L9).

xav (approved):
I'm approving if this is blocking you, otherwise, I'm not sure about `latest` since versioned tags haven't really been a problem and a generic one can be tricky sometimes.
Anyway, this will be overwritten with the migration.

geekbrother:
> Anyway, this will be overwritten with the migration.

@xav do you mean by migration of the future updates of the CD workflow to make it consistent across all services?
The reason for these changes is to control what version we are shipping to prod and distinguish to which environment to ship. At the moment it's kind of messy.

PR #231, opened 2023-10-02
fix: `notification` variable is encoded as a string, fix alert thresholds

### LINKED ISSUE #210: fix: alarm notifications disabled
https://github.com/WalletConnect/echo-server/pull/114/files#diff-301fdb9b35fae4b2dd8554c65981149e450f8ebfe68e76dfbf48e2b89f901486R7

geekbrother:
# Description

This PR fixes firing alerts for 5xx errors by making the following changes:
- Removing `jsonencode` from the local `notifications` variable that causes encoding json array as a string.
- Changing evaluation for the 5xx errors for every 5 minutes with 0 threshold. In this case, we can catch 500 errors even when the error appears and then disappears, so we can investigate it.
- Changing the threshold for an error count from 5 to 1 so we can catch and investigate a single 500 errors if they occur.
- Changing the policy for `no data state` from: `no_data` to keep the last state to provide consistent alarm firing in case the threshold was met and then no data was provided.

Resolves #210

## How Has This Been Tested?

Deployed from the PR branch to the staging environment, the alert was fired successfully and delivered to the opsgenie and slack channel when the 500 error count was > 1 for 5 minutes.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/monitoring/main.tf
+++ b/terraform/monitoring/main.tf
@@ -592,7 +592,7 @@ resource "grafana_dashboard" "at_a_glance" {
             }
           ],
           "executionErrorState" : "alerting",
-          "frequency" : "5m",
+          "frequency" : "1m",
           "handler" : 1,
           "name" : "${var.environment} Echo Server 5XX alert",
           "noDataState" : "keep_state",
--- a/terraform/monitoring/main.tf
+++ b/terraform/monitoring/main.tf
@@ -547,7 +547,7 @@ resource "grafana_dashboard" "at_a_glance" {
             {
               "evaluator" : {
                 "params" : [
-                  5
+                  1
                 ],
                 "type" : "gt"
               },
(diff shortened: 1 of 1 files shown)


chris13524 (approved):

xav (approved):

arein (approved):

terraform/monitoring/main.tf
```diff
@@ -592,12 +592,11 @@ resource "grafana_dashboard" "at_a_glance" {
             }
           ],
           "executionErrorState" : "alerting",
-          "for" : "5m",
-          "frequency" : "1m",
+          "frequency" : "5m",
```
chris13524: Why reduce frequency to every 5m? Maybe we can leave at 1m?

geekbrother: I think 5m for alerting frequency is enough to catch issues if they happen, but we can use 1m if we need a lower threshold. I've changed it to 1m.

PR #233, opened 2023-10-03
fix: changing client creation SQL query to handle id and token non-unique

### LINKED ISSUE #230: bug: duplicate key errors for the `clients.id` constraint
```
```

geekbrother:
# Description

This PR changes the SQL query in the client creation to handle behavior when the client tries to register with the same client id and device token as it has (they both are non-unique). This change fixes throwing 500 errors when this case happens.

Resolves #230 

## How Has This Been Tested?

It's covered by the functional unit tests for the client creation. The expected behavior is passing the CI gate.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/client.rs
+++ b/src/stores/client.rs
@@ -33,11 +33,15 @@ impl ClientStore for sqlx::PgPool {
 
         let mut transaction = self.begin().await?;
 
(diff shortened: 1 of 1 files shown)


geekbrother:
Closing this PR as it doesn't change anything in behavior according to the unit test in #234 which is run on the previous SQL query.

PR #234, opened 2023-10-04
test: adding `client_create_same_id_and_token` test

### LINKED ISSUE #230: bug: duplicate key errors for the `clients.id` constraint

geekbrother:
# Description

Adding the `client_create_same_id_and_token` functional test for the test client creation with the same `id` and `device_token` which are already present in the database.

Related to #230 

## How Has This Been Tested?

Covered by the CI Unit tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/tests/functional/stores/client.rs
+++ b/tests/functional/stores/client.rs
@@ -190,6 +190,70 @@ async fn client_upsert_id(ctx: &mut StoreContext) {
 
 #[test_context(StoreContext)]
 #[tokio::test]
+async fn client_create_same_id_and_token(ctx: &mut StoreContext) {
+    let client_id = format!("id-{}", gen_id());
+    let token = format!("token-{}", gen_id());
+
+    // Initial Client creation
+    ctx.clients
+        .create_client(TENANT_ID, &client_id, Client {
+            tenant_id: TENANT_ID.to_string(),
+            push_type: ProviderKind::Fcm,
+            token: token.clone(),
+        })
+        .await
+        .unwrap();
+    let insert_result = ctx.clients.get_client(TENANT_ID, &client_id).await.unwrap();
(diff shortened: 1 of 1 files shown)


xav (approved):

tests/functional/stores/client.rs
```diff
@@ -188,6 +188,70 @@ async fn client_upsert_id(ctx: &mut StoreContext) {
...
+            tenant_id: TENANT_ID.to_string(),
+            push_type: ProviderKind::Noop,
+            token: token.clone(),
+        })
+        .await
+        .unwrap();
+    let double_insert_result = ctx.clients.get_client(TENANT_ID, &client_id).await.unwrap();
+    assert_eq!(double_insert_result.token, token);
+    assert_eq!(double_insert_result.push_type, ProviderKind::Noop);
+
```
chris13524, line 247: Can we add (perhaps in another PR) a test that the notification still exists in the DB for this client? Also same for other tests where the client ID is the same

chris13524 (approved):

tests/functional/stores/client.rs
geekbrother: I think this case should not happen as we have [ON DELETE CASCADE in the notifications table](https://github.com/WalletConnect/echo-server/blob/3dde225228178c6198c89d674113c05ad4eb700b/migrations/1695381202_alter_notifications_constraint.sql#L6): notifications should be removed when the related client->id is removed.

chris13524: Yeah but I'm saying if the client_id is re-added, the notifications shouldn't be deleted.

geekbrother: Created #236 to track this comment.

PR #235, opened 2023-10-04
fix: adding `... FOR UPDATE` to lock the row while client creation

### LINKED ISSUE #230: bug: duplicate key errors for the `clients.id` constraint

geekbrother:
# Description

Adding an additional `SELECT ... FOR UPDATE` statement for explicitly locks the row and avoids the `unique constraint violation` errors while executing create client transactions in parallel.

Resolves #230

## How Has This Been Tested?

Passing functional CI unit tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/client.rs
+++ b/src/stores/client.rs
@@ -33,6 +33,13 @@ impl ClientStore for sqlx::PgPool {
 
         let mut transaction = self.begin().await?;
 
+        // Statement for locking the row. Issue #230
+        sqlx::query("SELECT * FROM public.clients WHERE id = $1 OR device_token = $2 FOR UPDATE")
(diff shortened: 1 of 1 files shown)


geekbrother:
The expected behavior is that when the row is locked by `SELECT ... FOR UPDATE` another concurrent transaction with `FOR UPDATE` will wait until the first one `COMMIT` and will not throw any errors in this case.
More about `FOR UPDATE` and concurrent transactions: https://chat.openai.com/share/9cf007cd-fb8f-4281-ba66-e5924d884ca2

chris13524 (approved):

PR #240, opened 2023-10-06
fix: use `axum-client-ip` to get the real client IP

### LINKED ISSUE #225: fix: geoip lookup errors for the grey IP networks
We have a lot of following errors:
```
2023-09-29T13:37:14.085015Z ERROR geoip lookup error: AddressNotFoundError: Address not found in database
```
reflecting that we are querying the geoip database with the non-existent (internal grey networks) IPs.
As a solution, we should use `X-Forwarded-For` header as a client IP source and skip the lookup in case it's absent.

geekbrother:
# Description

This PR adds using of `axum-client-ip` to get a real client IP which is included in the `X-Forwarded` by calling `SecureClientIp`. The AWS ELB[ uses X-Forwarded-for](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html) and [SecureClientIp sourced it for a real IP](https://github.com/imbolc/axum-client-ip/blob/a23f172ee4d01ead9ec25cedd3e6359ba63a03ad/src/secure.rs#L33).

Following changes are made:
 - `axum-client-ip` is added to the `Cargo.toml` as a dependency. We [already have it](https://github.com/WalletConnect/echo-server/blob/46d17972c3f83288fd849ca4c16ad15edb11caee/Cargo.lock#L607) as a relative dependency in the `Cargo.lock` so it will not add additional dependency weight.
 - `ConnectInfo` replaced with `SecureClientIp`.

Resolves #225

## How Has This Been Tested?

Successfully built and run with `analytics` feature enabled.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@ use {
         routing::{delete, get, post},
         Router,
     },
+    axum_client_ip::SecureClientIpSource,
     config::Config,
     hyper::http::Method,
     opentelemetry::{sdk::Resource, KeyValue},
@@ -210,7 +211,8 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) -> 
                     hyper::http::header::CONTENT_TYPE,
                     hyper::http::header::AUTHORIZATION,
                 ]),
-        );
+        )
+        .layer(SecureClientIpSource::RightmostXForwardedFor.into_extension());
 
     #[cfg(feature = "multitenant")]
     let app = {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,7 +1187,7 @@ checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "echo-server"
-version = "0.34.7"
+version = "0.35.0"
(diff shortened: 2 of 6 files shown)


chris13524 (approved):
LGTM

arein:
Nice vendor agnostic solution!

arein (approved):

src/lib.rs
```diff
@@ -210,7 +211,8 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
                     hyper::http::header::CONTENT_TYPE,
                     hyper::http::header::AUTHORIZATION,
                 ]),
-        );
+        )
+        .layer(SecureClientIpSource::ConnectInfo.into_extension());
```
chris13524:
```suggestion
        .layer(SecureClientIpSource::RightmostXForwardedFor.into_extension());
```

chris13524 (changes requested):

src/lib.rs
geekbrother: Good catch! Thanks, Chris. Applied it.

chris13524 (approved):

PR #241, opened 2023-10-09
fix: using relay public key from environment variable

### LINKED ISSUE #228: fix: hardcode relay public key from environment variable
To validate incoming webhooks for pushed messages (push_message.rs), the echo server checks that the request is signed by the relay's public key (relay/mod.rs). To get the public key, it requests this from https://relay.walletconnect.com/public-key. However, the [logs show](https://walletconnect.slack.com/archives/C03SPNLG39P/p1695330354814059) that occasionally bursts of hundreds of "IncompleteMessage" errors performing this request occur, which results in the relay needing to retry sending of the webhook, and delayed PNs.

The cause of these errors is unknown, and has [also happened](https://walletconnect.slack.com/archives/C03SMNKLPU0/p1694704186740059) when the notify server attempts to use HTTP RPC method of publishing to the relay. But this appears to be a network-layer related issue when the connection is closed before a response is fully sent.

Some potential solutions to this could involve handling the errors gracefully, and using a stale public key in this case. As well as using a async lock to only allow 1 request to be in-flight at a time.

However, due to the nature that this public key is fixed, is the same across all relay nodes, and the plan for how these keys are managed across many relay instances involves a [GitHub repo](https://walletconnect.slack.com/archives/C05ABTQSPFY/p1692365559949539?thread_ts=1690988442.746529&cid=C05ABTQSPFY) as a registry of valid public keys, it is preferable to scrap the current implementation of requesting from `/public-key` and instead simply use an environment variable to provide the fixed public key.
```[tasklist]
### Tasks
- [ ] Scrap requesting logic
- [ ] Obtain public key by requesting it from https://relay.walletconnect.com/public-key
- [ ] Add environment variable and hardcode the public key in Terraform config
```

geekbrother:
# Description

This PR changes the public key to be used from the environment variable instead of grabbing it from the relay URL.
The following changes (commits) are made:
 - Adding `public_key_url` to the environment variables and passing it from the GitHub secrets and to the terraform. Making it usable from the Config. ([commit](https://github.com/WalletConnect/echo-server/pull/241/commits/851ffe1e2b9a57ab9f5c61b7f4b29eb501be94ba)),
 - Scraping key requesting logic to use the key from config ([commit](https://github.com/WalletConnect/echo-server/pull/241/commits/411fef5a7e3954cc634b57f8a4469972509b09dd)),
 - Fixing unit tests to work with the updated logic. Extracting database URI to be passed as an environment variable. Removing hardcoded config variables to make the Config injectable. ([commit](https://github.com/WalletConnect/echo-server/pull/241/commits/3925136a05ce27e8e99ab841ff76a3f4b7fbeb81)).

Resolves #228 

## How Has This Been Tested?

Tested by functional unit tests in CI workflow

## Due Diligence

* [ ] Breaking change


.env.example
```diff
@@ -3,6 +3,9 @@ PUBLIC_URL=https://echo.walletconnect.com
 DATABASE_URL=postgres://user:pass@host:port/database
 DISABLE_HEADER=false
 
+# Public key can be obtained from the https://relay.walletconnect.com/public-key
```
chris13524:
```suggestion
# Public key can be obtained from https://relay.walletconnect.com/public-key
```

.env.multi-tenant-example
```diff
@@ -3,6 +3,9 @@ PUBLIC_URL=http://localhost:3000
 DATABASE_URL=postgres://user:pass@host:port/database
 LOG_LEVEL=debug,echo-server=debug
 
+# Public key can be obtained from the https://relay.walletconnect.com/public-key
```
chris13524:
```suggestion
# Public key can be obtained from https://relay.walletconnect.com/public-key
```

.env.single-tenant-example
```diff
@@ -3,6 +3,9 @@ PUBLIC_URL=http://localhost:3000
 DATABASE_URL=postgres://user:pass@host:port/database
 LOG_LEVEL=debug,echo-server=debug
 
+# Public key can be obtained from the https://relay.walletconnect.com/public-key
```
chris13524:
```suggestion
# Public key can be obtained from https://relay.walletconnect.com/public-key
```

tests/context/server.rs
```diff
@@ -22,19 +18,18 @@ pub struct EchoServer {
...
             log_level_otel: "info,echo-server=trace".into(),
             disable_header: true,
-            relay_url: "https://relay.walletconnect.com".into(),
+            relay_public_key: config_from_env.relay_public_key.clone(),
             validate_signatures: false,
-            database_url: DATABASE_URL.into(),
-            #[cfg(feature = "multitenant")]
-            tenant_database_url: TENANT_DATABASE_URL.into(),
+            database_url: config_from_env.database_url.clone(),
+            tenant_database_url: config_from_env.tenant_database_url.clone(),
```
chris13524: Looks like we lost the `#[cfg(feature = "multitenant")]`

```suggestion
            #[cfg(feature = "multitenant")]
            tenant_database_url: config_from_env.tenant_database_url.clone(),
```

src/config.rs
```diff
@@ -205,6 +212,18 @@ fn default_cors_allowed_origins() -> Vec<String> {
     vec!["*".to_string()]
 }
 
+fn default_jwt_secret() -> String {
+    "n/a".to_string()
+}
+
+fn default_analytics_export_bucket() -> String {
+    "example-bucket".to_string()
+}
```
chris13524: I don't think these make sense as defaults outside of a mocked test context. Some things are simply required and we should error if they aren't provided. Can we default these elsewhere, inside the tests perhaps? E.g. by constructing `Config` manually with test config, rather than using `envy::from_env::<Config>()` with default values.

tests/context/server.rs
```diff
@@ -22,19 +18,18 @@ pub struct EchoServer {
 pub enum Error {}
 
 impl EchoServer {
-    pub async fn start() -> Self {
+    pub async fn start(config_from_env: &Config) -> Self {
```
chris13524: Why did you refactor all of this? I don't see any newly added tests.

.env.example
geekbrother: Fixed, thanks!

tests/context/server.rs
geekbrother: Thanks for catching it, but It was not added because of `TENANT_DATABASE_URL` [was hardcoded ](https://github.com/WalletConnect/echo-server/blob/591effe09eb1ca8e0b63011f785a06c0b4ca9c86/tests/context/mod.rs#L14) before and to use it conditionally more refactoring is needed (like we opening both connections for a pg pool, etc.). 
As this PR became pretty big for a review, this refactoring should go as a follow-up.

src/config.rs
geekbrother: Makes sense. I've removed recently added default functions and using of `from_env::<Config>`. Now it's [added from the certain env variable](https://github.com/WalletConnect/echo-server/pull/241/commits/b4f8e427fac33e1342516347b433d76be67ca4af#diff-0b4c46909397317bf44e57ab8ed53fd89084fbb45b73d1958415ef3034e0bafaR43) when constructing the Config for test.

tests/context/server.rs
geekbrother: We are [constructing the Config in one place](https://github.com/WalletConnect/echo-server/pull/241/files#diff-0b4c46909397317bf44e57ab8ed53fd89084fbb45b73d1958415ef3034e0bafaR33-R86) at the upper level now, so it makes sense just to inject it and remove the creation of the sub-Config from the start function and make it lean.

chris13524 (approved):

PR #246, opened 2023-10-15
fix: fixing apns keys encoding

geekbrother:
# Description

To fix the multi-tenant apns update bug (where apns key is always wrong), the following changes and fixes are made:

Commit [163155f](https://github.com/WalletConnect/echo-server/pull/246/commits/163155f77eea6892f1f2424c32093b9623a2b388) fix:
- Removing [base64 encoding from the p8 ke](https://github.com/WalletConnect/echo-server/blob/80e3bf5c7ab5537c31bbe2c034f805e77f368ef8/src/handlers/update_apns.rs#L140)y as it's in already base64 key format (PEM),
- Added [missed decoding from base64 for the p12 key](https://github.com/WalletConnect/echo-server/blob/80e3bf5c7ab5537c31bbe2c034f805e77f368ef8/src/handlers/update_apns.rs#L182-L196),
- Adding values extraction case when the password is not provided (key without password), as [it was missed](https://github.com/WalletConnect/echo-server/blob/80e3bf5c7ab5537c31bbe2c034f805e77f368ef8/src/handlers/update_apns.rs#L51-L95) and always returning `Invalid Multipart Body` when the password was not provided,
- Replacing `BufReader` with `Cursor` as we don't need to create a buffer and can use just a cursor.

Commit [917c36a](https://github.com/WalletConnect/echo-server/pull/246/commits/917c36aa3b5898477cbf732ccf1c900027068335) test:
- Functional unit test `tenant_update_apns_valid_token` was created to cover testing apns update using the `reqwest` post request with the multipart form with valid apns p8 key from the environment variable,
- Adding required CI variables passing to run `tenant_update_apns_valid_token` unit test with valid apns p8 key from GitHub secrets.

Resolves #244 

## How Has This Been Tested?

It was tested by the CI unit test `tenant_update_apns_valid_token` from this PR.
The expected behavior is passing the `tenant_update_apns_valid_token` from the CI gate.

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

PR #253, opened 2023-10-18
fix: revert to the base64 encoded p8 key

### LINKED ISSUE #244: bug: APNS certificate validation errors
There is an error when using p12 certificate without a password for the APNS.
Context: https://walletconnect.slack.com/archives/C03RU4F6P8A/p1697184620615089

The following error occurs when validating the p8 certificate:
```
[Expecting: ANY PRIVATE KEY"](library: "PEM routines", function: "get_name", reason: "no start line", file: "../crypto/pem/pem_lib.c", line: 745, data: "Expecting: ANY PRIVATE KEY")
```

geekbrother:
# Description

This PR reverting back to double base64 encoding for the p8 apns key.
As the p8 key is already in the PEM (base64) format we used the base64 encoded key before #246 and to prevent confusion and make a proper migration - reverting back to the base64 encoded PEM format.

Additional tracing logs are also added in this PR.

Resolves #244 

## How Has This Been Tested?

p8 key validation is tested by [tenant_update_apns_valid_token](https://github.com/WalletConnect/echo-server/blob/main/tests/functional/multitenant/apns.rs#L62) unit test in the CI gate.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/tenant.rs
+++ b/src/stores/tenant.rs
@@ -19,7 +19,7 @@ use {
     chrono::{DateTime, Utc},
     serde::{Deserialize, Serialize},
     sqlx::{Executor, PgPool},
-    std::io::BufReader,
+    tracing::{info, instrument},
 };
 
 #[cfg(any(debug_assertions, test))]
@@ -178,6 +178,7 @@ impl Tenant {
         }
     }
 
+    #[instrument(skip_all, fields(tenant_id = %self.id, provider = %provider.as_str()))]
     pub fn provider(&self, provider: &ProviderKind) -> Result<Provider> {
         if !self.providers().contains(provider) {
             return Err(ProviderNotAvailable(provider.into()));
@@ -196,12 +197,11 @@ impl Tenant {
                         &self.apns_topic,
                     ) {
                         (Some(certificate), Some(password), Some(topic)) => {
+                            info!("apns certificate (p12) provider is matched");
                             let decoded =
                                 base64::engine::general_purpose::STANDARD.decode(certificate)?;
-                            let mut reader = BufReader::new(&*decoded);
-
                             let apns_client = ApnsProvider::new_cert(
-                                &mut reader,
+                                &mut &mut std::io::Cursor::new(decoded),
                                 password.clone(),
                                 endpoint,
                                 topic.clone(),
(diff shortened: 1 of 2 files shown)


src/stores/tenant.rs
```diff
@@ -218,9 +219,15 @@ impl Tenant {
                         &self.apns_team_id,
                     ) {
                         (Some(topic), Some(pkcs8_pem), Some(key_id), Some(team_id)) => {
-                            let mut reader = BufReader::new(pkcs8_pem.as_bytes());
+                            info!("apns token (p8) provider is matched");
+                            // For supporting the legacy format of the p8 key which is double base64 encoded
```
chris13524: If it's "double" base64 encoded, then won't decoding always pass?

chris13524 (changes requested):

src/stores/tenant.rs
geekbrother: The updated p8 key format in the database is plain text like:
```
-----BEGIN PRIVATE KEY-----
...base64 encoded key...
-----END PRIVATE KEY-----
```
which is actually the content of the p8 key file.

The legacy format was that content but encoded in base64 with the `BEGIN/END PRIVATE KEY` header and footer.
So the new format should not be decoded successfully (because of the header and footer) and we should pass the plain text instead in `unwrap_or`.

We should encode p12 because it's content is in bytes, but skip for further "double" encoding p8.

geekbrother: Resolving this as we've decided to go back to the base64 encoded PEM key for p8.

src/stores/tenant.rs
```diff
@@ -218,9 +218,12 @@ impl Tenant {
                         &self.apns_team_id,
                     ) {
                         (Some(topic), Some(pkcs8_pem), Some(key_id), Some(team_id)) => {
-                            let mut reader = BufReader::new(pkcs8_pem.as_bytes());
+                            info!("apns token (p8) provider is matched");
+                            let p8_token = base64::engine::general_purpose::STANDARD
+                                .decode(pkcs8_pem)
+                                .expect("p8 token is not a valid base64 string");
```
chris13524: Use `?` instead of `.expect()`

chris13524 (changes requested):

src/stores/tenant.rs
geekbrother: When using `?` here we will miss the context, if we use `expect` we can pass additional context right to the error, so the error will be like `p8 token is not a valid base64 string: Bad character at $%!`.

geekbrother: Removing it because this can't be handled properly in this behavior ((

chris13524 (approved):

PR #257, opened 2023-10-19
chore: distinguish bad device token errors

### LINKED ISSUE #256: chore: apns and fcm bad token responses are considered as an internal server errors
We have a lot of `apns(unregistered)` HTTP 500 errors. An `unregistered device token` is not an internal server error and should be distinguished from it. `BadDeviceToken` error should be used for such cases.

geekbrother:
# Description

We have a lot of `apns(unregistered)` HTTP 500 errors. An `unregistered device token` is not an internal server error and should be distinguished from it. `BadDeviceToken` error should be used for such cases.

1. This PR changes the `BadDeviceToken` error type to have a string argument that reflects a certain device token error.
2. Distinguish apns and fcm responses from bad device tokens and other types that should throw 500 (Internal Server Error).

Resolves #256 

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -25,8 +25,8 @@ pub enum Error {
     #[error(transparent)]
     Prometheus(#[from] prometheus_core::Error),
 
-    #[error("invalid device token provided")]
-    BadDeviceToken,
+    #[error("Bad device token error: {0}")]
+    BadDeviceToken(String),
 
     #[error(transparent)]
     Apns(#[from] a2::Error),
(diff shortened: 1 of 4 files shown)


xav (commented):
Why are we returning a 400 when a dependency fails? The client has nothing to do with it, wouldn't it be better with a 503 instead?

xav:
> Why are we returning a 400 when a dependency fails? The client has nothing to do with it, wouldn't it be better with a 503 instead?

That would also mean we should stop monitoring 5xx and only monitor 500 to see actual server errors.

geekbrother:
> That would also mean we should stop monitoring 5xx and only monitor 500 to see actual server errors.

As a possible solution, we can check if the error is `Unregistered` and throw another return code such as 4xx and throw 500 otherwise to catch all non `Unregistered` errors. Because we don't want to monitor all client's unregistered token errors and have an alert for them.

xav:
> > That would also mean we should stop monitoring 5xx and only monitor 500 to see actual server errors.
> 
> As a possible solution, we can check if the error is `Unregistered` and throw another return code such as 4xx and throw 500 otherwise to catch all non `Unregistered` errors. Because we don't want to monitor all client's unregistered token errors and have an alert for them.

That would be nice since it would let us still see the actual errors instead of hiding everything under 400s

geekbrother:
> That would be nice since it would let us still see the actual errors instead of hiding everything under 400s

We have a specific error for the `Unregistered` in A2, I'll just extract it and throw an error only for that case. Thanks @xav!

xav (approved):

chris13524 (approved):

PR #261, opened 2023-10-21
fix: adding jwt verification to the apns and fcm update handler

### LINKED ISSUE #262: bug: missing jwt validation in apns and fcm update handler
Missing JWT validation check in apns and fcm update handlers.
Updates can be done without providing the JWT token.

geekbrother:
# Description

This PR adds missing JWT verification to the:
- update apns handler,
- update fcm handler.

Resolves #262

## How Has This Been Tested?

Functional tests for valid apns and fcm update [fails without providing jwt](https://github.com/WalletConnect/echo-server/actions/runs/6598301402/job/17926141443?pr=261#step:11:871).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/update_apns.rs
+++ b/src/handlers/update_apns.rs
@@ -1,18 +1,21 @@
 use {
     crate::{
         error::{Error, Error::InvalidMultipartBody},
+        handlers::validate_tenant_request,
         increment_counter,
+        request_id::get_req_id,
         state::AppState,
         stores::tenant::{TenantApnsUpdateAuth, TenantApnsUpdateParams},
     },
     axum::{
         extract::{Multipart, Path, State},
+        http::HeaderMap,
         Json,
     },
     base64::Engine,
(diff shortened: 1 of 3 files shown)


arein:
Looks like a breaking change? How many users will we break

geekbrother:
> Looks like a breaking change? How many users will we break

Is it used only by our cloud app? If so, we already have the JWT header for these requests in our cloud app [here](https://github.com/WalletConnect/cloud-app/blob/18704f9f68022804b864503d8acc044689699be8/src/utils/NextApiUtil.ts#L255) and [here](https://github.com/WalletConnect/cloud-app/blob/18704f9f68022804b864503d8acc044689699be8/src/utils/NextApiUtil.ts#L282). That's why my assumption that the JWT check was just forgotten.

chris13524 (approved):

arein (approved):

PR #274, opened 2023-11-08
fix: idempotency per client

### LINKED ISSUE #273: fix: idempotency per client

chris13524:
# Description

If multiple clients are subscribed to the same topic and are registered to receive push notifications, this won't work because the idempotency logic is designed to prevent the same message ID from being notified multiple times.

This behavior was actually flakey due to the idempotency logic itself not being robust against race conditions, but this is a separate issue. 

Context: https://walletconnect.slack.com/archives/C044SKFKELR/p1699446818332569?thread_ts=1699442986.858579&cid=C044SKFKELR

Resolves #273

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,11 @@ jobs:
             args: --features multitenant,analytics,geoblock,functional_tests
             cache: { sharedKey: "tests" }
             rustc: stable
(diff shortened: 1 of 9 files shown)


geekbrother (approved):
Good catch! Thanks for fixing it.

PR #277, opened 2023-11-13
chore: style sql

chris13524:
# Description

Make SQL style consistent. While it's generally a bad idea to change existing migrations, these should be functionally equivalent and won't be rerun anyway because the migration ID is what determines if the migration was already run or not.

This PR matches notify-server styling and has a companion PR [here](https://github.com/WalletConnect/notify-server/pull/207).

This style matches what I've generally seen in the Postgres documentation:
- lowercase type names
- capital constraint names (sometimes I see lowercase, but uppercase seems more common)
- capital `ARRAY`
- opening parenthesis on same line
- 4 space indents
- space between the constraint name and the columns it applies to e.g. `UNIQUE (device_token)`

Also as a personal preference:
- no `IF NOT EXISTS` for table creation as these are migrations, and shouldn't exist anyway (and if they do, there is a potential for being out of sync)
- generally have newlines between each statement
- trailing newlines at end of file

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


arein (approved):
Sweet - FYI [in other repos we use `sqlfluff` for this](https://github.com/WalletConnect/data-lake/blob/4c169356f60fe122a7a6e9e2470a29ef4eea0557/.github/workflows/dbt-compile.yml#L36-L41) - it's an awesome tool

geekbrother (approved):

chris13524:
@arein nice! We should pull in something like this. Would be nice if it had a brew install though

PR #279, opened 2023-11-14
feat(decrypted_notify): adding `always_raw` for the client registration

geekbrother:
# Description

This PR adds `always_raw` field to the clients database table and to the client register endpoint as an optional field with the `false` as default to follow the [specs for the decrypted notifications](https://github.com/WalletConnect/walletconnect-specs/pull/147/files#diff-25490e5dda4bcd46802fca4131439d45ab85cf408683b76c7604735dfe656edcR76).

Related to #278 

## How Has This Been Tested?

Updated functional tests.

## Review stack 🏗️
1 - Adding always_raw for the client registration #279 <-
2 - Update message payload and pass raw message #281

## Due Diligence

* [ ] Breaking change
* [ ] Requires a documentation update
* [ ] Requires a e2e/integration test update

Issue #278, opened 2023-11-14
feat: following the decrypted notifications specs


geekbrother:
This issue tracks changes on the echo-server for the [decrypted notification specs](https://github.com/WalletConnect/walletconnect-specs/pull/147).

chris13524:
Already implemented

PR #281, opened 2023-11-16
feat(decrypted_notify): update message payload and pass raw message

geekbrother:
# Description

This PR adds `topic`, `tag` and `message` fields for the send notification request payload to follow the [specs for the decrypted notifications](https://github.com/WalletConnect/walletconnect-specs/pull/147/files#diff-25490e5dda4bcd46802fca4131439d45ab85cf408683b76c7604735dfe656edcR76).

The handler modification checks if the client was registered (have) `always_raw` as `true` and then handles the message as an encrypted with the payload from the `message` field. Otherwise, handle the payload as usual to support backward compatibility.

Related to #278

## How Has This Been Tested?

* Current clients support is tested by the current unit/integration tests (passing current tests on new changes),
* New clients with `always_raw = true` and push message with the new required fields are tested by the new [test_push_always_raw test](https://github.com/WalletConnect/echo-server/pull/281/files#diff-513a9ee6d914ac38439771435145221152a2aeafff7c1a41aba304842354ec79R193).

## Review stack 🏗️
1 - Adding always_raw for the client registration #279
2 - Update message payload and pass raw message #281 <-

## Due Diligence

* [ ] Breaking change


src/providers/apns.rs
```diff
@@ -69,33 +67,45 @@ impl PushProvider for ApnsProvider {
...
         // TODO tidy after https://github.com/WalletConnect/a2/issues/67 is closed
-        let result = match payload.is_encrypted() {
-            true => {
-                let mut notification_payload = a2::DefaultNotificationBuilder::new()
-                    .set_content_available()
-                    .set_mutable_content()
-                    .set_title("You have new notifications. Open to view")
-                    .build(token.as_str(), opt);
+        // Sending `always_raw` encrypted message
+        let result = if body.topic.is_some() && body.tag.is_some() && body.message.is_some() {
```
chris13524: It should only do this when `always_raw=true` not when these flags are `Some`

src/providers/fcm.rs
```diff
@@ -27,18 +27,35 @@ impl FcmProvider {
...
     ) -> crate::error::Result<()> {
-        let s = span!(tracing::Level::DEBUG, "send_fcm_notification");
-        let _ = s.enter();
-
         let mut message_builder = MessageBuilder::new(self.api_key.as_str(), token.as_str());
 
-        let result = if payload.is_encrypted() {
-            message_builder.data(&payload)?;
+        // Sending `always_raw` encrypted message
+        let result = if body.topic.is_some() && body.tag.is_some() && body.message.is_some() {
```
chris13524: It should only do this when `always_raw=true` not when these flags are `Some`

chris13524 (changes requested):

src/providers/apns.rs
geekbrother: Do these fields are required when the `always_raw=true` by the spec? So we know that the client is `always_raw=true` when they are present?

geekbrother: Btw, I think [these two](https://github.com/WalletConnect/echo-server/pull/281/files#diff-b90d908d3c5121717714b75753e7f32aec4b723e226dfa3c270e3510fe9d63beR63-R64) should become optional, because of the new client.

chris13524: > Do these fields are required when the always_raw=true by the spec?

Yes, you can't support clients with `always_true` if the relay hasn't been upgraded yet to support. In practice this usually won't happen except during initial deployment, and we most likely will deploy the relay update first anyway.

> So we know that the client is always_raw=true when they are present?

No, the relay has no knowledge of `always_true` value, and relay upgraded with [this PR](https://github.com/WalletConnect/rs-relay/pull/1285) will always send the fields.

> Btw, I think [these two](https://github.com/WalletConnect/echo-server/pull/281/files#diff-b90d908d3c5121717714b75753e7f32aec4b723e226dfa3c270e3510fe9d63beR63-R64) should become optional, because of the new client.

Not a bad idea and would enable smoother migration later if we remove these flags from the relay, so makes sense. If `always_true=true` then require the new fields, and if `always_true=false` then require the old fields. Could use `#[serde(flatten)]` potentially to avoid checking more than 2 `Option`s

geekbrother: Thanks for clarifying! 

> Btw, I think [these two](https://github.com/WalletConnect/echo-server/pull/281/files#diff-b90d908d3c5121717714b75753e7f32aec4b723e226dfa3c270e3510fe9d63beR63-R64) should become optional, because of the new client.

> Not a bad idea and would enable smoother migration later if we remove these flags from the relay, so makes sense. If always_true=true then require the new fields, and if always_true=false then require the old fields. Could use #[serde(flatten)] potentially to avoid checking more than 2 Options

I think making fields optional should be a follow-up.

geekbrother: > It should only do this when always_raw=true not when these flags are Some

Fixed. [Check for required fields](https://github.com/WalletConnect/echo-server/pull/281/files#diff-b90d908d3c5121717714b75753e7f32aec4b723e226dfa3c270e3510fe9d63beR196-R206) is in the handler and [in the test](https://github.com/WalletConnect/echo-server/pull/281/files#diff-513a9ee6d914ac38439771435145221152a2aeafff7c1a41aba304842354ec79R207-R226) now.

src/providers/fcm.rs
```diff
@@ -27,18 +27,36 @@ impl FcmProvider {
...
+                topic: body.topic.unwrap(),
+                tag: body.tag.unwrap(),
+                message: body.message.unwrap(),
+            })?;
+
+            // Must set priority=high and content-available=true on data-only messages or
+            // they don't show unless app is active
+            // https://rnfirebase.io/messaging/usage#data-only-messages
+            message_builder.priority(Priority::High);
+            message_builder.content_available(true);
```
chris13524: Would be nice to make this code-block DRY, e.g. with a local function

src/providers/fcm.rs
```diff
@@ -27,18 +27,36 @@ impl FcmProvider {
...
-
         let mut message_builder = MessageBuilder::new(self.api_key.as_str(), token.as_str());
 
-        let result = if payload.is_encrypted() {
-            message_builder.data(&payload)?;
+        // Sending `always_raw` encrypted message
+        let result = if always_raw {
+            info!("Sending raw encrypted message");
+            message_builder.data(&RawMessagePayload {
+                topic: body.topic.unwrap(),
```
chris13524: These unwraps scare me a little. I wonder if this could be refactored somehow so that the state of these values is tied to `always_raw`? E.g. `if let Some(...) = raw_values`?

chris13524 (approved):
LGTM, but please do not merge until https://github.com/WalletConnect/rs-relay/pull/1285 is merged so we can ensure there are no backwards incompatibilities.

src/providers/fcm.rs
geekbrother: Done.

src/providers/fcm.rs
geekbrother: That makes sense, thx!
Now it's done with the [check](https://github.com/WalletConnect/echo-server/pull/281/files#diff-3f172be4d81562a96e145c96cc9cf4abcf9ec3db3c65d476d58f37b397133981R80-R86) and throw an appropriate error.

chris13524: I meant a way to avoid errors altogether. Like instead of passing `always_raw`, you check these in the calling function, and then pass the raw values through. So then you just check if there are raw values, rather than if `always_raw=true` and then asserting they are Somes a second time

src/handlers/push_message.rs
```diff
@@ -175,18 +178,52 @@ pub async fn handler_internal(
         )
     })?;
 
+    let cloned_body = body.clone();
+    let push_message = if client.always_raw {
```
geekbrother, line 179: Probably this condition or our spec description should be changed because in our spec we [require the tag to be presented](https://github.com/WalletConnect/walletconnect-specs/pull/147/files#diff-25490e5dda4bcd46802fca4131439d45ab85cf408683b76c7604735dfe656edcR82) as well if `always_raw == false` to filter by tag and send clear text notification for some of the tags. I think we should fix this in a spec and do tag filtering when the `always_raw == true`. cc @chris13524

chris13524: > When set to `false`, cleartext notifications will be sent for certain messages based on `tag`

This filtering is [performed in the relay](https://github.com/WalletConnect/rs-relay/pull/1285/files#diff-7644ae4ab3ab50067abfcd3582ce60dcf0eade53bc256dea0acf373f4bdbfb57R554), and was already implemented. So I don't think anything needs to be done here.

Cargo.toml
```diff
@@ -66,7 +66,7 @@ fcm = "0.9"
 ed25519-dalek = "2.0.0-rc.2"
 
 # JWT Authentication
-relay_rpc = { git = "https://github.com/WalletConnect/WalletConnectRust.git", rev = "ced99e7"}
+relay_rpc = { git = "https://github.com/WalletConnect/WalletConnectRust.git", rev = "4ee9007"}
```
xav: Must we use hashes on repos we control?
It would be much better to make a release and use the tag in our dependencies.

xav (approved):

Cargo.toml
chris13524: I switched it 👍

PR #286, opened 2023-11-17
feat: decrypted notifications analytics

### LINKED ISSUE #282: Analytics for always_raw preference
Export metrics on client registrations that enable `always_raw` so we know % of clients that upgraded to use this new feature

https://github.com/WalletConnect/echo-server/blob/ea1401e818aae6cf7ab0b0819cf4358f2469608b/src/handlers/register_client.rs#L124

chris13524:
# Description

Adds `always_raw` field to client info export, set to the value the client provided.

Adds `always_raw` to message export, which is set to the same value of the client, or null if the client cannot be found.

Makes `encrypted` and `flags` optional, since these are optionally provided from the relay now.

Adds `tag` field which is provided by the relay, and is technically optional during initial rollout.

Resolves #282 

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change


src/analytics/message_info.rs
```diff
@@ -11,9 +11,10 @@ pub struct MessageInfo {
     pub client_id: Arc<str>,
     pub topic: Arc<str>,
     pub push_provider: Arc<str>,
-    pub always_encrypted: bool,
-    pub encrypted: bool,
-    pub flags: u32,
+    pub always_raw: Option<bool>,
+    pub tag: Option<u32>,
+    pub encrypted: Option<bool>,
```
geekbrother, line 16: Non-blocking. It's a bit confusing to have `always_raw` and `encrypted`. The value from `encrypted` comes from `body.legacy` maybe we should use `legacy_encrypted`? So it will not confuse the data team as well.

geekbrother (approved):
lgtm, except a comment.

src/analytics/message_info.rs
chris13524: We can't change the scheme since there is already data that's been exported using it. We'd have to wipe the data or adjust queries to exclude the old schema, which can be done but in this case doesn't seem necessary

PR #287, opened 2023-11-28
feat(o11y): makes the alarm less noisy

arein:
# Description

Our downstream services are not expected to have 100% uptime which this alarm expected. So this ups the limit a bit to reduce noise.

inb4: it's an `OR` operator so if any of the sources are > 5 it fires an alarm.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change


geekbrother (approved):

chris13524:
Should have a process to review any stray 5xx's so we can continue to nail and eliminate them

PR #293, opened 2023-12-13
fix: adding the pg_lock for the token

### LINKED ISSUE #292: fix: concurrent client creation with the same token and different client id
When concurrent client creation with the same token and different client ids occurs we have a `unique constraint violation error`.
This error occurs because we are using only the `id` as a key for the [pg_advisory_xact_lock when creating a lock in the transaction](https://github.com/WalletConnect/echo-server/blob/3fca011fa9b2e398198b5b5621260d0a82385f54/src/stores/client.rs#L40C18-L40C18).

geekbrother:
# Description

This PR adds an additional `pg_advisory_xact_lock` based on the device token to prevent data race when concurrent client creations occur with the same token id and different client IDs, because at the moment only the client ID is used for the lock. In such cases, we have an `unique constraint violation` error for the device token because we have [a parallel insert](https://github.com/WalletConnect/echo-server/blob/3fca011fa9b2e398198b5b5621260d0a82385f54/src/stores/client.rs#L51).

The `pg_advisory_xact_lock`s are released at the [end of the transaction](https://github.com/WalletConnect/echo-server/blob/3fca011fa9b2e398198b5b5621260d0a82385f54/src/stores/client.rs#L70).

Resolves #292

## How Has This Been Tested?

Current functional [database tests](https://github.com/WalletConnect/echo-server/actions/runs/7194773344/job/19596015598?pr=293#step:11:803).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/client.rs
+++ b/src/stores/client.rs
@@ -36,16 +36,16 @@ impl ClientStore for sqlx::PgPool {
         let mut transaction = self.begin().await?;
 
         // Statement for locking based on the client id to prevent an issue #230
-        sqlx::query("SELECT pg_advisory_xact_lock(abs(hashtext($1::text)))")
-            .bind(id)
-            .execute(&mut transaction)
-            .await?;
-
-        // Statement for locking based on the token to prevent an issue #292
-        sqlx::query("SELECT pg_advisory_xact_lock(abs(hashtext($1::text)))")
-            .bind(client.token.clone())
-            .execute(&mut transaction)
-            .await?;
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
LGMT. Can we do both of these in the same statement to save RTT?

geekbrother:
> LGMT. Can we do both of these in the same statement to save RTT?

Good catch! Yes, we can do `SELECT pg_lock.., pg_lock;`. Tested it in the local Postgres, and it works as expected.

PR #296, opened 2024-01-25
fix: using client id `pg_advisory_xact_lock` lock when inserting notification

### LINKED ISSUE #295: bug: key `client_id` is not present in table `clients`
We have rare database error messages containing the following error:

```
Store(Database(Database(PgDatabaseError { severity: Error, code: "23503", message: "insert or update on table \"notifications\" violates foreign key constraint \"fk_notifications_client_id\"", detail: Some("Key (client_id)=(...) is not present in table \"clients\"."), hint: None, position: None, where: None, schema: Some("public"), table: Some("notifications"), column: None, data_type: None, constraint: Some("fk_notifications_client_id"), file: Some("ri_triggers.c"), line: Some(2509), routine: Some("ri_ReportViolation") })))
```

We have a [client_id related constraint in the database schema](https://github.com/WalletConnect/echo-server/blob/main/migrations/1695381202_alter_notifications_constraint.sql) that prevents inserting the record to the notifications table without the related client id.

At this error, we are trying to insert the notification without the corresponding client id in the `clients` table.

When investigating the issue by connecting to the database and inspecting `clients` and `notifications` tables, the inspection reveals that the client with the corresponding id is present in the table for the id that caused the error.

We have two different endpoints for the client creation and the notification publishing. When the client creation calls there is [INSERT and DELETE](https://github.com/WalletConnect/echo-server/blob/d77fa504ca6098a73126cdc78b7f1a8040e7706f/src/stores/client.rs#L50-L57) into the clients table.

The constraint error occurs as a race condition when two endpoints call in parallel and in the first we are deleting the client id for that we are inserting the new notification from the other notification endpoint before the new row is inserted.

geekbrother:
# Description

This PR adds using of the pg xact lock related to the client id to prevent the race condition described in #295.
We are using this [xact lock with the client id as a key when the client creation](https://github.com/WalletConnect/echo-server/blob/d77fa504ca6098a73126cdc78b7f1a8040e7706f/src/stores/client.rs#L42) calls to avoid the race condition.

Resolves #295

## How Has This Been Tested?

Unit test for the `create_or_update_notification`.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/notification.rs
+++ b/src/stores/notification.rs
@@ -50,6 +50,13 @@ impl NotificationStore for sqlx::PgPool {
         client_id: &str,
         payload: &PushMessageBody,
     ) -> stores::Result<Notification> {
+        let mut transaction = self.begin().await?;
+
+        sqlx::query("SELECT pg_advisory_xact_lock(abs(hashtext($1::text)))")
+            .bind(client_id)
+            .execute(&mut transaction)
+            .await?;
+
         let res = sqlx::query_as::<sqlx::postgres::Postgres, Notification>(
             "
             INSERT INTO public.notifications (id, tenant_id, client_id, last_payload)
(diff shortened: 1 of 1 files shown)


chris13524 (approved):
Good fix! Still don't like the use of locks here, but we weren't able to figure out any alternatives 😭

PR #298, opened 2024-01-31
fix: tenants update on conflict handling

geekbrother:
# Description

We have rare errors where the client tries to create tenants with the same `id` that causes the following errors:
```
"duplicate key value violates unique constraint \"tenants_pkey\"", detail: Some("Key (id)=(xxx) already exists."
```

This PR adds:
* Handling of `ON CONFLICT` while creating the new tenant.
* Adds missing updates of `updated_at` field on any updates of the record.

## How Has This Been Tested?

Functional tests for multitenant.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/tenant.rs
+++ b/src/stores/tenant.rs
@@ -302,7 +302,11 @@ impl TenantStore for PgPool {
     #[instrument(skip(self))]
     async fn create_tenant(&self, params: TenantUpdateParams) -> Result<Tenant> {
         let res = sqlx::query_as::<sqlx::postgres::Postgres, Tenant>(
-            "INSERT INTO public.tenants (id) VALUES ($1) RETURNING *;",
+            "INSERT INTO public.tenants (id)
+            VALUES ($1)
+            ON CONFLICT (id)
+            DO UPDATE SET updated_at = NOW()
+            RETURNING *",
         )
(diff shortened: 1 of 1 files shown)


src/stores/tenant.rs
```diff
@@ -314,7 +318,7 @@ impl TenantStore for PgPool {
     #[instrument(skip(self))]
     async fn update_tenant(&self, id: &str, params: TenantUpdateParams) -> Result<Tenant> {
         let res = sqlx::query_as::<sqlx::postgres::Postgres, Tenant>(
-            "UPDATE public.tenants SET id = $2 WHERE id = $1 RETURNING *;",
+            "UPDATE public.tenants SET id = $2, updated_at = NOW() WHERE id = $1 RETURNING *;",
```
chris13524, line 321: What is the purpose even of this function? It looks like this is never called anyway. Should just remove as `id` shouldn't be updated.

chris13524 (approved):
Perfect.

src/stores/tenant.rs
geekbrother: It was added a long ago, probably the author expect to use it somehow. I'll remove this rudiment.

PR #307, opened 2024-03-05
fix: scale down

chris13524:
# Description

- Downscales RDS and ECS to minimal configuration in non-prod environment.
- Downscales RDS min capacity from 2 to 1 since this server is not heavy on the database and usage seems stable at exactly or lower than 2.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -103,7 +103,7 @@ module "database_cluster" {
   enabled_cloudwatch_logs_exports = ["postgresql"]
 
   serverlessv2_scaling_configuration = {
-    min_capacity = 2
+    min_capacity = local.environment == "prod" ? 1 : 0.5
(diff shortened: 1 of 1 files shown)


nopestack (approved):

terraform/main.tf
```diff
@@ -103,7 +103,7 @@ module "database_cluster" {
   enabled_cloudwatch_logs_exports = ["postgresql"]
 
   serverlessv2_scaling_configuration = {
-    min_capacity = 2
+    min_capacity = local.environment == "prod" ? 1 : 0.5
     max_capacity = 10
```
chris13524:
```suggestion
    max_capacity = 20
```

PR #309, opened 2024-03-13
fix: increasing Postgres connection pool size

geekbrother:
# Description

This PR increases the Postgres connection pool size from `5` to `30` to handle request spikes properly.
During the request spikes there are `Store(Database(PoolTimedOut)` errors, that should be fixed by increasing the pool size. The RDS state at this moment remains stable (ACU <= 30%), so it should be handled without scaling.

## How Has This Been Tested?

Unit testing.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,8 @@ pub mod state;
 pub mod stores;
 pub mod supabase;
 
+const PG_CONNECTION_POOL_SIZE: u32 = 30;
+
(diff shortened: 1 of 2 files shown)


xDarksome (commented):
Why is the change under `/tests`?

geekbrother:
> Why is the change under `/tests`?

last `git push` was missed, I didn't think you would review so fast 😅👍

xDarksome (approved):

PR #317, opened 2024-04-16
feat: refactored cloud auth

chris13524:
# Description

Uses the new Cloud authentication mechanism introduced in [this PR](https://github.com/WalletConnect/cloud-app/pull/1094). [Slack conversation](https://walletconnect.slack.com/archives/C03SPNLG39P/p1713271392759799)

Instead of a Supabase user JWT being provided to Push Server, Push Server authenticating this JWT with a shared secret from Cloud, and then Push Server checking with Push Server if the user has access to the tenant... now the authorization of the user's access to the project/tenant is completely offloaded to Cloud and all Push Server does now is checks a JWT signature. This JWT is now signed by Cloud app with a different shared secret and has a TTL of 30s rather than being the same one used for all user authentication in Cloud.

Resolves that Cloud currently is not able to configure Push Server due to a breaking change that was overlooked.

Remaining work:
- [x] Update `JWT_SECRET` to new secret
- [x] Remove `CLOUD_API_SECRET` secret

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/update_apns.rs
+++ b/src/handlers/update_apns.rs
@@ -123,22 +123,13 @@ pub async fn handler(
     headers: HeaderMap,
     mut form_body: Multipart,
 ) -> Result<Json<UpdateTenantApnsResponse>, Error> {
-    // Ensure tenant real
-    let _existing_tenant = state.tenant_store.get_tenant(&id).await?;
-
     // JWT verification
     #[cfg(feature = "cloud")]
-    let jwt_verification_result = validate_tenant_request(
-        &state.registry_client,
-        &state.gotrue_client,
-        &headers,
-        id.clone(),
-        None,
-    )
-    .await;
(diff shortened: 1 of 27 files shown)


src/handlers/update_apns.rs
```diff
@@ -149,6 +140,9 @@ pub async fn handler(
         return Err(e);
     }
 
+    // Ensure tenant real
+    let _existing_tenant = state.tenant_store.get_tenant(&id).await?;
```
chris13524, line 144: Moving after, auth comes first

nopestack (approved):

Cali93 (approved):
Nice one! One thing that we could look into in a separate PR is to using staging registry for project id validation, like Notify Server does.

PR #327, opened 2024-05-15
feat: per-IP rate limiting

### LINKED ISSUE #304: IP-based rate limiting

geekbrother:
# Description

This PR implements the simple per-IP rate limiting. The rate limiter is implemented in an Axum middleware using the in-memory `Moka` caching. The default requests per second limit is 10 req/sec which should be enough for the legit users and cut flood requests. Only the `X-Forwarded` IP is used to prevent blocking the Load Balancer IP.

Resolves #304 

## How Has This Been Tested?

* Manual testing,
* Basic [flood integration test](https://github.com/WalletConnect/push-server/pull/327/files#diff-29ee1e76a6651d33c3e1a5117e948675ccda1e60165c885696a9accdc1bf0bb9R72).

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/integration/integration.test.ts
+++ b/integration/integration.test.ts
@@ -69,4 +69,37 @@ describe('Echo Server', () => {
       expect(status).toBe(200)
     })
   })
+  describe('Middlewares', () => {
+    const httpClient = axios.create({
+      validateStatus: (_status) => true,
+    })
+
+    // Simulate flood of requests and check for rate-limited responses
+    it('Rate limiting', async () => {
+      const url = `${BASE_URL}/health`
+      const requests_to_send = 100;
+      const promises = [];
+      for (let i = 0; i < requests_to_send; i++) {
+        promises.push(
+          httpClient.get(url)
+        );
+      }
+      const results = await Promise.allSettled(promises);
+      
+      let ok_statuses_counter = 0;
+      let rate_limited_statuses_counter = 0;
+      results.forEach((result) => {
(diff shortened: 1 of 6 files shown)


nopestack (approved):
i'm curious about the memory increase using this implementation will incur in the server. Is there a limit to how much memory is allocated to the rate limiter?

geekbrother:
> i'm curious about the memory increase using this implementation will incur in the server. Is there a limit to how much memory is allocated to the rate limiter?

Thanks for the comment! The memory is used to store entries of IP addresses and counters. The IP address is coming from `X-Forwarded`, so from the LB. AWS and other major infrastructures prevent spoofing IP addresses at the router levels (BGP etc.) so if the attacker even has 1M IP addresses it will use a few of megabytes memory. We are not pre-allocating memory in this case, making the size dynamic.

chris13524 (commented):
I think this PR applies the same 10 req/s rate limit to the `push_message` handler which needs a different rate limit since this endpoint is used constantly ([looks like about 20 req/s](https://g-e66b9a1a24.grafana-workspace.eu-central-1.amazonaws.com/d/prod-relay-ng/prod-relay-ng?orgId=1&from=now-15m&to=now)) by a single group of relay servers vs individual clients.

Testing it manually it seems to work, but I'm guessing this is luck because there are multiple replicas of both push and relay that would pull the req/s under the threshold.

PR #328, opened 2024-05-16
fix: applying rate limiting middleware to all except push endpoints

geekbrother:
# Description

This PR applying the rate-limiting middleware to all endpoints *except* message push because it's used by the Relay and validated by the signature.

Also, this PR fixes the bug of double applying the `geoblock` and `with state` middlewares.

## How Has This Been Tested?

* Current integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -235,62 +235,65 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) -> 
                 ),
             );
 
-        let app = Router::new()
-            .route("/health", get(handlers::health::handler))
-            .nest("/tenants", tenancy_routes)
+        Router::new()
+            .route("/health", get(handlers::health::handler).layer(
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -236,61 +236,64 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
...
+                ),
             )
+            // Rate limiting middleware is not applying to push_handler because it is used by the relay
             .route(
                 "/:tenant_id/clients/:id",
                 post(handlers::push_message::handler),
             )
             .layer(global_middleware);
-
-        let app = if let Some(geoblock) = state_arc.geoblock.clone() {
```
geekbrother, line 255: Removing this part as we are applying the geoblock and state later and not depending on the multitenant feature. This is a fix to double applying the geoblock and state.

chris13524 (approved):
Thanks!

PR #331, opened 2024-05-17
fix: rate limit 10 req/min

chris13524:
# Description

[Slack conversation](https://walletconnect.slack.com/archives/C058RS0MH38/p1715961479758389?thread_ts=1715900528.828279&cid=C058RS0MH38)

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/middleware/rate_limit.rs
+++ b/src/middleware/rate_limit.rs
@@ -11,8 +11,6 @@ use std::{net::IpAddr, sync::Arc};
 use tokio::time::Duration;
 use tracing::error;
 
-pub const MAX_REQUESTS_PER_SEC: u32 = 10;
-
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


src/middleware/rate_limit.rs
```diff
@@ -11,8 +11,6 @@ use std::{net::IpAddr, sync::Arc};
 use tokio::time::Duration;
 use tracing::error;
 
-pub const MAX_REQUESTS_PER_SEC: u32 = 10;
```
chris13524, line 14: Didn't rename to PER_MIN and removed instead since the desire to change rate limit is more than just changing this constant; we may change the window too

geekbrother (approved):
lgtm!

PR #332, opened 2024-05-17
fix: client registrations chart

chris13524:
# Description

This chart is confusing since it shows registrations over the past hour, but per sec or per rate interval is really what we care about.

Also it was an up/down counter which was strange and the Grafana didn't like with that (it was warning). Moving to only-up counter to get us closer to a req/s i.e. understanding of load vs the arbitrary amount of registrations we currently have.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/terraform/monitoring/main.tf
+++ b/terraform/monitoring/main.tf
@@ -150,71 +150,6 @@ resource "grafana_dashboard" "at_a_glance" {
         "type" : "stat"
       },
       {
-        "datasource" : {
-          "type" : "prometheus",
-          "uid" : grafana_data_source.prometheus.uid
-        },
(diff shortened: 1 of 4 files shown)


geekbrother (approved):
Thanks for these changes!

terraform/monitoring/main.tf
```diff
@@ -149,71 +149,6 @@ resource "grafana_dashboard" "at_a_glance" {
         "title" : "Notifications per Hour",
         "type" : "stat"
       },
-      {
```
chris13524, line 152: Redundant chart

PR #334, opened 2024-05-20
fix: upgrade otel

chris13524:
# Description

There were some dependencies on old versions of OTEL which were making implementation of Histograms a little tricky. Removing these old versions.

This also removes OTEL tracing which I think doesn't even work anyway.

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,6 @@ use {
     config::Config,
     hyper::http::Method,
     middleware::rate_limit::rate_limit_middleware,
-    opentelemetry::{sdk::Resource, KeyValue},
     sqlx::{
         postgres::{PgConnectOptions, PgPoolOptions},
         ConnectOptions,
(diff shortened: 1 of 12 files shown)


geekbrother (approved):

src/lib.rs
```diff
@@ -264,9 +255,7 @@ pub async fn bootstap(mut shutdown: broadcast::Receiver<()>, config: Config) ->
 
     #[cfg(not(feature = "multitenant"))]
     let app = Router::new()
-        .route("/health", get(handlers::health::handler).layer(
-            axum::middleware::from_fn_with_state(state_arc.clone(), rate_limit_middleware),
-        ))
+        .route("/health", get(handlers::health::handler))
```
chris13524, line 258: Removes rate limits on `/health` as 1) they aren't needed and 2) they fail to resolve the `Forwarded-For` header anyway since it's called internally

PR #337, opened 2024-05-20
fix: improve create client performance

chris13524:
# Description

Currently we are deleting and re-inserting the record, and it turns out the delete operation is quite expensive compared to the insert. Refactoring the queries to only do the operations necessary: insert or update while still locking the record for update.

[Slack conversation](https://walletconnect.slack.com/archives/C03SPNLG39P/p1716246429442999?thread_ts=1716218673.594629&cid=C03SPNLG39P)

## How Has This Been Tested?

Not tested

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/stores/client.rs
+++ b/src/stores/client.rs
@@ -131,7 +131,11 @@ impl ClientStore for sqlx::PgPool {
             notification_query_builder.push(" and tenant_id = ");
             notification_query_builder.push_bind(existing_client.tenant_id);
             let notification_query = notification_query_builder.build();
+            let start = Instant::now();
             self.execute(notification_query).await?;
+            if let Some(metrics) = metrics {
+                metrics.postgres_query("create_client_delete_notification", start);
+            }
 
             let query = "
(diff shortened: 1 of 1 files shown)


nopestack:
> Currently we are deleting and re-inserting the record
Interesting decision. Was there a particular reason this way implemented that way?

nopestack (approved):

chris13524:
> Interesting decision. Was there a particular reason this way implemented that way?

It was the culmination of multiple iterations of new queries not susceptible to concurrency errors. I think we went with DELETE because it was easier than handling all of the insert or update cases (of which this PR now implements directly).

geekbrother (approved):
lgtm, thanks! One comment: did we miss the [transaction between](https://github.com/WalletConnect/push-server/pull/337/files#diff-b6bb7dd16c438697d757ed306cadbf12c75e4dff9e97a0bbbe33b639da82cd30L72) select and updates?

PR #341, opened 2024-06-04
fix: concurrent client registration errors

chris13524:
# Description

Fixes race condition where two INSERTs for the same client ID are made due to clients calling this endpoint in rapid succession.

Alternative would be to return a 409 conflict or assume success when INSERT gives a conflict, but this 1) hides potential errors and 2) doesn't live up to the purest form of idempotency.

## How Has This Been Tested?

New test

## Due Diligence

* [ ] Breaking change


arein (approved):
nice!

geekbrother (approved):
Thanks for fixing this!

PR #343, opened 2024-06-20
fix(apns): proper handling of the wrong CA in APNS certificate

geekbrother:
# Description

This PR adds an additional catching and handling of the `received fatal alert: UnknownCA` APNS client error.
The same action was added to disable the tenant in case of the unknown CA certificate error as for the expired APNS certificate since the certificate can't be used to send the push notification until it is updated.

## How Has This Been Tested?

Not tested.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -184,6 +184,9 @@ pub enum Error {
     #[error("Expired APNs certificate")]
     ApnsCertificateExpired,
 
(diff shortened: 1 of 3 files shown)


chris13524 (approved):
Thanks!!

PR #346, opened 2024-07-24
fix: adding check for the maximum payload size

geekbrother:
# Description

This PR implements a check for the maximum payload size, limiting the size to 4Kb, and throws an appropriate message if exceeding. 
The reason for limiting the body payload size is the FCM support payload up to 4Kb. 
This should fix the following error:
```
FcmV1(UnknownHttpResponse { status: 400, body: Ok("{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Android message is too big\",\n    \"status\": \"INVALID_ARGUMENT\",\n    \"details\": [\n      {\n        \"@type\": \"type.googleapis.com/google.rpc.BadRequest\"\n      },\n      {\n        \"@type\": \"type.googleapis.com/google.firebase.fcm.v1.FcmError\",\n        \"errorCode\": \"INVALID_ARGUMENT\"\n      }\n    ]\n  }\n}\n") })
```

## How Has This Been Tested?

Not tested, but should be covered by the unit and integration tests.

## Due Diligence

* [ ] Breaking change

### DIFF
--- a/src/handlers/push_message.rs
+++ b/src/handlers/push_message.rs
@@ -168,6 +168,18 @@ pub async fn handler_internal(
         )
     })?;
 
+    // Check request body size shouldn't exceed 4Kb (FCM limitations)
+    if serde_json::to_string(&body)
+        .map_err(|e| (Error::InternalSerializationError(e), None))?
+        .len()
+        > 4096
+    {
+        return Err((
+            Error::PayloadTooLarge("Request body size should not exceed 4Kb".to_string()),
+            None,
+        ));
+    };
+
     let cloned_body = body.clone();
     let push_message = if client.always_raw {
         if let Some(body) = body.raw {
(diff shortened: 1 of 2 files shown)


src/handlers/push_message.rs
```diff
@@ -168,6 +168,18 @@ pub async fn handler_internal(
...
     })?;
 
+    // Check request body size shouldn't exceed 4Kb (FCM limitations)
+    if serde_json::to_string(&body)
+        .map_err(|e| (Error::InternalSerializationError(e), None))?
+        .len()
+        > 4096
+    {
+        return Err((
+            Error::PayloadTooLarge("Request body size should not exceed 4Kb".to_string()),
```
nopestack, line 178: Why not make that inner string part of the display implementation provided by `thiserror` and you just embed the size of the payload in kb?

nopestack (approved):

chris13524 (commented):
This is not really an appropriate fix IMO. This is blaming the relay for the payload being too big but the relay is just forwarding messages so it doesn't address the root issue.

Why is the payload too big?

PR #386, opened 2026-09-10
fix(ci): unbreak auto_deploy — container build, clippy, FCM test, staging CD vars

geekbrother:
Unbreaks `main`'s `auto_deploy` ([run 34381941835](https://github.com/reown-com/push-server/actions/runs/34381941835)), which fails in three independent ways. None is caused by a commit — the same failures are in the pre-#384 run [34351560435](https://github.com/reown-com/push-server/actions/runs/34351560435), and `run-cd` has been skipped in every retained run, so nothing has deployed from `main`.

### 1. `build-container` (both Dockerfiles)

```
error: failed to compile `cargo-chef v0.1.78`
  feature `edition2024` is required
  ... not stabilized in this version of Cargo (1.79.0 (ffa9cf99a 2024-06-03))
```

`rust:buster` is frozen at Rust 1.79 because Debian buster is EOL, while unpinned `cargo install cargo-chef` resolves the newest release, whose `clap_derive` needs edition 2024 — a moving dependency against a frozen toolchain. Both files move to bookworm (build and runtime stages together, so the binary and its libssl stay consistent) and `cargo install` gains `--locked`.

### 2. Clippy (all three feature variants)

9 errors under `-D warnings` from lints in the newer floating stable: `unused import: debug` in `src/handlers/delete_fcm_v1.rs`, and 8× `useless_borrows_in_formatting` (a redundant `&` in a `format!` arg) in `src/error.rs` and `src/handlers/push_message.rs`.

Fixing those surfaced two dead-code errors queued behind the lib failure: an empty unreferenced `pub enum Error {}` in `tests/context/server.rs`, removed; and the whole `context` harness being unused under default features, now gated behind `functional_tests` to match its only consumer, `mod functional`.

### 3. `tenant_update_fcm_bad`

Google decommissioned the legacy FCM HTTP API in June 2024. Probed directly:

```
POST https://fcm.googleapis.com/fcm/send  →  HTTP 404
```

`update_fcm.rs` validates a legacy key with a dry-run send and maps only `FcmError::Unauthorized` to `BadFcmApiKey`; a 404 hits the `_ => Ok(())` arm, so every key is accepted. The assertion cannot hold while legacy FCM is offered.

Failing closed is the obvious fix and is wrong: `tenant_update_fcm_valid`, `tenant_enabled_providers` and `tenant_delete` each post a key from `ECHO_TEST_FCM_KEY` and assert success, passing only because validation falls open. Rejecting on any error turns one red test into four and makes legacy key registration impossible in production.

The test is therefore ignored with its reason, and the validation site records what it no longer does: any key is accepted and stored, a suspended tenant is restored on it, and `src/providers/fcm.rs` sends through the same dead endpoint — so legacy FCM pushes are already failing in production. **Dropping legacy FCM in favour of the v1 path is the real fix and a product decision, deliberately not made here.**

### 4. CD would have failed next

`run-cd` calls `cd.yml`, whose apply steps set `TF_VAR_jwt_secret`, `TF_VAR_relay_public_key` and `TF_VAR_image_version` — all ignored by TFC remote-execution runs, so the apply would fail on the two required variables just as the plan did. `image_version` is the subtle one: unset, `coalesce()` in `main.tf` falls back to the latest release instead of the image `auto_deploy` just pushed under `github.sha`. The staging job now injects them via `apply.auto.tfvars.json` and applies inline, matching `ci_terraform.yml` and keys-server's `sub-infra-apply.yml`.

> [!IMPORTANT]
> **`deploy-infra-prod` still needs the identical change** and is not included here. It is skipped by `auto_deploy` (`deploy_to_prod: false`), so it does not block this run; manual prod deploys stay broken exactly as they are today until it lands.

### Verification

- `cargo clippy` with `-D warnings` for `--tests`, `--features=multitenant --tests`, and `--all-features --all-targets`: clean
- `cargo fmt --check`: clean
- `cargo test --features <the CI set> --test integration -- --list --ignored`: exactly one ignored test
- **full `docker build -f Dockerfile .`, all stages: exit 0** — chef, plan, release compile with `multitenant,analytics,cloud`, and the `bookworm-slim` runtime
- bookworm carries the build packages (`protobuf-compiler`, `lld`, `llvm`) and runtime `libssl3`

Note that no PR check builds the container: `build-container` exists only in `auto_deploy.yml` (`push: main`) and `release.yml` (`workflow_dispatch`). That is why the Dockerfile was verified locally, and arguably why this went unnoticed — a build-only PR job would have caught it at the commit that broke it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -69,15 +69,29 @@ jobs:
         with:
           environment: "staging"
 
+      # echo-server-* are TFC remote-execution workspaces, which ignore runner-side
+      # TF_VAR_*; a *.auto.tfvars.json file ships with the uploaded configuration and is
+      # read by the run. image_version matters here: left unset, coalesce() in main.tf
+      # falls back to the latest release instead of the image just built.
+      - name: Configure Terraform Variables
+        working-directory: terraform
+        env:
+          JWT_SECRET: ${{ secrets.STAGING_JWT_SECRET }}
+          RELAY_PUBLIC_KEY: ${{ secrets.RELAY_PUBLIC_KEY }}
+          IMAGE_VERSION: ${{ inputs.image_tag }}
+        run: |
+          jq -n --arg jwt "$JWT_SECRET" --arg relay "$RELAY_PUBLIC_KEY" --arg version "$IMAGE_VERSION" \
+            '{jwt_secret: $jwt, relay_public_key: $relay, image_version: $version}' > apply.auto.tfvars.json
+
+      # Inline rather than WalletConnect/actions/terraform/apply: that action passes
+      # -var-file, which is not how variables reach a remote run. Matches the plan job
+      # in ci_terraform.yml and keys-server's sub-infra-apply.yml.
       - name: Deploy Terraform to Staging
         id: tf-apply
-        uses: WalletConnect/actions/terraform/apply/@1.0.3
         env:
-          TF_VAR_jwt_secret: ${{ secrets.STAGING_JWT_SECRET }}
-          TF_VAR_image_version: ${{ inputs.image_tag }}
-          TF_VAR_relay_public_key: ${{ secrets.RELAY_PUBLIC_KEY }}
-        with:
-          environment: "staging"
+          TF_WORKSPACE: staging
+          TF_IN_AUTOMATION: "true"
+        run: terraform -chdir=terraform apply -auto-approve -no-color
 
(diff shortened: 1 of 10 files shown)


arminx86-77616c (approved):
Thanks for putting the vars into `apply.auto.tfvars.json`

